Issue #1, opened 2018-09-29
Not support "ignore" in JSON config of data source


zhoney:
There might be columns useless to construct vertex/edge, hugegraph-loader should provide a method to ignore them rather than pre-processing them by users before insert.

PR #2, opened 2018-09-29
add ignore mechanism to data source JSON config file

zhoney:
implement: #1

Change-Id: I915bb5e0dfc35c57a902b1d440fe4dcde65ec683


javeme:
@zhoney synch with icode

PR #24, opened 2018-11-28
Fix edge parser error check property key when customized id

Linary:
Change-Id: If53feef102ecd66b78dc0f2ac2f83c4664b20a29


src/main/java/com/baidu/hugegraph/loader/parser/EdgeParser.java
```diff
@@ -106,6 +103,8 @@ private Object buildVertexId(VertexLabel vertexLabel,
                     return fieldValue;
                 }
             } else {
+                String key = this.source.mappingField(fieldName);
```
javeme: check value is a number if IdStrategy.CUSTOMIZE_NUMBER

PR #26, opened 2018-11-28
Support to skip the line start with specified symbols

Linary:
Change-Id: Id5a664a3d0c09bac7cb3761862ab69bfd8cca5dd


src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -67,4 +73,8 @@ public String delimiter() {
     public String charset() {
         return this.charset;
     }
+
+    public Set<String> commentSymbols() {
+        return this.commentSymbols;
```
javeme: unmodify

PR #28, opened 2018-11-29
Support customized date format

Linary:
Implement #27

Change-Id: Ib00a1b6d53dee82c7be7cb8420662787707161d5


src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -31,6 +31,7 @@
 public class FileSource implements InputSource {
 
     private static final String DEFAULT_CHARSET = "UTF-8";
+    private static final String DEFAULT_DATE_FORMAT = "yyyy-MM-dd-HH-mm-ss";
```
javeme: use space between date and time: "yyyy-MM-dd HH-mm-ss"

PR #33, opened 2018-12-29
Support load with a specified directory

Linary:
Implement #29

Change-Id: Id3d9385b52484a30d6e71652ef1b55c40979dd88


src/main/java/com/baidu/hugegraph/loader/reader/file/FileReader.java
```diff
@@ -97,17 +100,29 @@ public boolean hasNext() {
 
     @Override
     public void close() throws IOException {
-        this.reader.close();
+        // pass
```
javeme: maybe BufferedReaderWrapper is not closed if readNextLine() has never been called.

PR #43, opened 2019-03-18
Support config regex to skip specified line

Linary:


src/main/java/com/baidu/hugegraph/loader/reader/file/AbstractFileReader.java
```diff
@@ -123,16 +123,17 @@ protected Line fetch() {
...
             return this.fetch();
         } else {
             return this.parser.parse(rawLine);
         }
     }
 
-    private boolean isCommentLine(String line) {
-        return this.source.commentSymbols().stream().anyMatch(line::startsWith);
+    private boolean needSkipLine(String line) {
+        String regex = this.source.skipLineRegex();
```
javeme: set default regex if not set by config

src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -45,16 +43,15 @@
     private String charset;
     @JsonProperty("date_format")
     private String dateFormat;
+    @JsonProperty("skip_line_regex")
```
javeme: skipped_line_regex

src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -90,13 +89,12 @@ public String dateFormat() {
         return this.dateFormat;
     }
 
-    public Compression compression() {
-        return this.compression;
+    public String skippedLineRegex() {
+        return this.skippedLineRegex;
```
javeme: return different default value based on date_format

javeme: so this will skip empty line

PR #46, opened 2019-03-19
Add an option 'token' for authentication

Linary:
Change-Id: Id2d748f83ce3c52dc961ba32b044dff6e40b6615


src/main/java/com/baidu/hugegraph/loader/executor/LoadOptions.java
```diff
@@ -51,6 +51,11 @@
                description = "The port of HugeGraphServer")
     public int port = 8080;
 
+    @Parameter(names = {"--token"}, arity = 1,
+               validateWith = {PositiveValidator.class},
```
javeme: may not be PositiveValidator?

Issue #59, opened 2019-04-18
Support value mapping


Linary:
Assume we have a file in format```UserID::Gender::Age::Occupation::Zip-code```

- Occupation is chosen from the following choices:

	*  0:  "other" or not specified
	*  1:  "academic/educator"
	*  2:  "artist"
	*  3:  "clerical/admin"
	*  4:  "college/grad student"
	*  5:  "customer service"
	*  6:  "doctor/health care"
	*  7:  "executive/managerial"
	*  8:  "farmer"
	*  9:  "homemaker"
	* 10:  "K-12 student"
	* 11:  "lawyer"
	* 12:  "programmer"
	* 13:  "retired"
	* 14:  "sales/marketing"
	* 15:  "scientist"
	* 16:  "self-employed"
	* 17:  "technician/engineer"
	* 18:  "tradesman/craftsman"
	* 19:  "unemployed"
	* 20:  "writer"

Loader should support map number to real occupation

PR #67, opened 2019-04-22
Support value mapping

Linary:
Change-Id: I39cad735f176cc772dd51c48f4d61537bbcf1c51

Implement #59


src/main/java/com/baidu/hugegraph/loader/builder/EdgeBuilder.java
```diff
@@ -101,16 +105,23 @@ private Object buildVertexId(VertexLabel vertexLabel,
                  * just return when id strategy is CUSTOMIZE_NUMBER
                  */
                 if (idStrategy.isCustomizeString()) {
-                    String id = (String) fieldValue;
+                    E.checkArgument(mappedValue instanceof String,
+                                    "If there is no value mapping, the field " +
+                                    "value must be String, otherwise it must " +
```
javeme: The field value must be String if there is no value mapping, the same as the value after mapping...

src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -63,7 +63,12 @@ public SourceType type() {
 
     @Override
     public void check() throws IllegalArgumentException {
-        // pass
+        String elemDelimiter = this.listFormat().elemDelimiter();
+        if (this.delimiter != null && this.delimiter.equals(elemDelimiter)) {
+            throw new IllegalArgumentException(String.format(
+                      "The delimiter '%s' can't equal with list elem " +
```
javeme: The delimiters of fields and list elements can't be the same

src/main/java/com/baidu/hugegraph/loader/source/file/FileSource.java
```diff
@@ -66,8 +66,8 @@ public void check() throws IllegalArgumentException {
         String elemDelimiter = this.listFormat().elemDelimiter();
         if (this.delimiter != null && this.delimiter.equals(elemDelimiter)) {
             throw new IllegalArgumentException(String.format(
-                      "The delimiter '%s' can't equal with list elem " +
-                      "delimiter '%s'", this.delimiter, elemDelimiter));
+                      "The delimiter of fields(%s) and list elements(%s) " +
```
javeme: "The delimiters"
and prefer to use E.checkArgument()

PR #80, opened 2019-06-16
Enhance summary info

imbajin:
1. add parserSuccess to comfirm data integrity
2. add parserSpeed to get performance results directly
3. record the import speed of each data source in the log for comparison

### DIFF
--- a/src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
+++ b/src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
@@ -62,7 +62,6 @@ public class HugeGraphLoader {
 
     private long parseFailureNum = 0L;
     private long parseSuccessNum = 0L;
-    private long parseSuccessPerSource = 0L;
 
     public static void main(String[] args) {
         HugeGraphLoader loader = new HugeGraphLoader(args);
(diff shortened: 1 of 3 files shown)


Linary:
split into two commits, one is summary info, another is http-connection

src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -192,6 +207,7 @@ private void loadVertex(VertexBuilder builder) {
             try {
                 Vertex vertex = builder.next();
                 batch.add(vertex);
+                this.parseSuccessPerSource++;
```
javeme: prefer to let loadVertex() return parseSuccessPerSource, and don't define parseSuccessPerSource as field

src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -178,8 +191,10 @@ private LoadSummary loadVertices() {
         LoadSummary summary = new LoadSummary("vertices");
         Duration duration = Duration.between(beginTime, endTime);
         summary.parseFailure(this.parseFailureNum);
+        summary.parserSuccess(this.parseSuccessNum);
```
Linary: parserSuccess -> parseSuccess

PR #99, opened 2019-11-06
Fix list or set property update failed

Linary:
Change-Id: Ia4d307776b86c6bcbe2b06957b4cf4cb921d9c3d


src/main/java/com/baidu/hugegraph/loader/util/DataTypeUtil.java
```diff
@@ -158,6 +158,9 @@ private static Object parseMultiValues(Object values, DataType dataType,
                      "Only accept AbstractSource when parse multi values, " +
                      "but got '%s'", source.getClass().getName());
         ListFormat listFormat = ((AbstractSource) source).listFormat();
+        E.checkArgument(listFormat != null,
+                        "The list_format must be setted when parse " +
```
javeme: must be set

PR #117, opened 2019-12-10
Let single edge insert support options.checkVertex

Linary:
Change-Id: Id559e0d1e5ce4dc6f54b68646e1883a4fa6fb9e9


src/main/java/com/baidu/hugegraph/loader/task/InsertTask.java
```diff
@@ -82,6 +82,20 @@ public LoadMetrics metrics() {
         return this.batch;
     }
 
+    @SuppressWarnings("unchecked")
+    protected void addBatch(ElemType type, List<Record<GE>> batch,
+                            boolean check) {
```
javeme: boolean checkVertex
also update updateBatch

PR #120, opened 2019-12-25
Split num threads for batch and single insert

Linary:
Change-Id: Ib00182949aa3ce74a5c7170f537e8463b03d0d7e


src/main/java/com/baidu/hugegraph/loader/task/TaskManager.java
```diff
@@ -70,19 +70,19 @@ public TaskManager(LoadContext context) {
...
     }
 
     private int batchSemaphoreNum() {
-        return 1 + this.options.numThreads;
+        return 1 + this.options.batchInsertThreads;
     }
 
     private int singleSemaphoreNum() {
-        return 2 * this.options.numThreads;
+        return 2 * this.options.singleInsertThreads;
```
javeme: make single task queue 2x threads number?

PR #141, opened 2020-03-10
Ensure primary key columns valid or participate field mapping

Linary:
Change-Id: I7bd1cc6c6333032b0ba1f2460a6fe25696fab420


src/main/java/com/baidu/hugegraph/loader/builder/ElementBuilder.java
```diff
@@ -228,6 +228,10 @@ protected static String spliceVertexId(VertexLabel vertexLabel,
         StringBuilder vertexId = new StringBuilder();
         StringBuilder vertexKeysId = new StringBuilder();
         for (Object value : primaryValues) {
+            E.checkArgument(value != null,
+                            "There are some columns haven't correspond to " +
+                            "primary keys, please modify header column or " +
+                            "field_mapping");
```
javeme: please show which primaryKey

PR #151, opened 2020-04-23
Some improve for adapting hubble

Linary:
Change-Id: Id6a75b45cd00f9e9285bcd54ecf50342d0dafe69

### DIFF
--- a/src/test/resources/parquet_compress_file/core-site.xml
+++ b/src/test/resources/parquet_compress_file/core-site.xml
@@ -1,0 +1,6 @@
+<configuration>
+    <property>
+        <name>fs.defaultFS</name>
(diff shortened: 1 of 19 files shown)


src/test/resources/parquet_compress_file/core-site.xml
```diff
@@ -0,0 +1,6 @@
+<configuration>
+    <property>
+        <name>fs.defaultFS</name>
+        <value>hdfs://localhost:8020</value>
```
houzhizhen, line 4: Is it going be work at running time? If it work at running time, it will prevent the override setting.

Linary: It just worked in tests

PR #164, opened 2020-07-23
Support timestamp as date format

Linary:
Change-Id: I1b100b52f762046e3bdfa2c1df44b66a223758b3

### DIFF
--- a/src/main/java/com/baidu/hugegraph/loader/util/HugeClientHolder.java
+++ b/src/main/java/com/baidu/hugegraph/loader/util/HugeClientHolder.java
@@ -20,6 +20,7 @@
 package com.baidu.hugegraph.loader.util;
 
 import com.baidu.hugegraph.driver.HugeClient;
+import com.baidu.hugegraph.driver.HugeClientBuilder;
 import com.baidu.hugegraph.exception.ServerException;
 import com.baidu.hugegraph.loader.constant.Constants;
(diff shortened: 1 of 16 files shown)


src/main/java/com/baidu/hugegraph/loader/util/HugeClientHolder.java
```diff
@@ -35,20 +35,15 @@ public static HugeClient create(LoadOptions options) {
         } else {
             address = options.host + ":" + options.port;
         }
+        String username = (options.username != null) ? options.username :
+                          options.graph;
```
javeme, line 39: prefer align like:
```
String username = options.username != null ?
                  options.username : options.graph;
```

src/test/java/com/baidu/hugegraph/loader/test/functional/LoadTest.java
```diff
@@ -40,7 +40,7 @@
     protected static final String CONFIRM_CLEAR = "I'm sure to delete all data";
     protected static final String URL = String.format("http://%s:%s",
                                                       SERVER, PORT);
-    protected static final HugeClient CLIENT = new HugeClient(URL, GRAPH);
+    protected static final HugeClient CLIENT = HugeClient.builder(URL, GRAPH).build();
```
javeme, line 43: exceed 80 chars

PR #166, opened 2020-07-29
Skip vertex whose primary values are all empty

Linary:
Implement #165

Change-Id: Ie9e1cc1f4141f9f63f57734faf36caefacef9691


src/main/java/com/baidu/hugegraph/loader/builder/ElementBuilder.java
```diff
@@ -243,10 +244,16 @@ private String spliceVertexId(VertexLabel vertexLabel,
...
                                                   Constants.TARGET_LIST);
             }
+            if (!pkValue.isEmpty()) {
+                allEmpty = false;
+            }
             vertexKeysId.append(pkValue);
             vertexKeysId.append("!");
         }
-
+        if (allEmpty) {
```
javeme, line 253: how to deal with partial empty

Linary: partial empty is valid

PR #169, opened 2020-09-24
Set loading mode before load and reset after load

Linary:
Change-Id: Ic8edcea08178db295313c9c40e173a0e6438c6e8


src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -342,6 +343,7 @@ private void stopThenShutdown() {
                 this.manager.shutdown();
             }
         } finally {
+            this.context.unsetLoadingMode();
```
javeme, line 346: add try-finally, call this.context.close() in finally block

javeme (approved):

zhoney (approved):

PR #194, opened 2021-02-05
Use array to optimize access perf in parse stage

Linary:
Change-Id: I681255011c76067cef465f4687c0092dd08d2102


src/main/java/com/baidu/hugegraph/loader/builder/EdgeBuilder.java
```diff
@@ -167,4 +176,36 @@ public void extractProperties(Map<String, Object> rawKeyValues) {
...
+        idx = 0;
+        for (String field : this.mapping.targetFields()) {
+            for (int pos = 0; pos < names.length; pos++) {
+                String name = names[pos];
+                if (field.equals(name)) {
+                    index.targetIndexes[idx++] = pos;
+                }
+            }
+        }
+        this.vertexIdsIndex = index;
```
javeme, line 209: prefer return index and assign in build() method for more readable

javeme (approved):

zhoney (approved):

PR #206, opened 2021-04-23
Load vertices before edges when check-vertex is true

Linary:


src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -315,6 +316,28 @@ private void handleParseFailure() {
         }
     }
 
+    private List<InputStruct> splitStructsIfNeeded(List<InputStruct> structs) {
```
javeme, line 319: update comment of executeParseTask: "Execute parse task sync"

src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -315,6 +316,28 @@ private void handleParseFailure() {
...
+        if (!this.context.options().checkVertex) {
+            return structs;
+        }
+        LOG.info("Load option check-vertex is true, force to load vertices " +
```
javeme, line 323: Forced to load vertices before edges since set option check-vertex=true

src/main/java/com/baidu/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -187,6 +187,7 @@ private void loadInputs() {
 
     private void load(List<InputStruct> structs) {
         // Load input structs one by one
+        structs = this.splitStructsIfNeeded(structs);
```
javeme, line 190: add log before and after load, and print which file and label

javeme, line 190: must wait for all vertices are complete(include insert) before load edges

zhoney (approved):

PR #266, opened 2022-04-20
chore: use .asf.yaml for apache workflow rule & add RAT in action

imbajin:
TODO:
- [ ] enable the required CI before release
- [ ] using the latest commitID in CI for server
- [x] enable rebase option for co-coding
- [x] enable rat check in action 

Do in next release: (1.x)
- [ ] merge all loader/client/hubble/tools action files into one so that we could reuse the env (could refer the [skywalking-ci](https://github.com/apache/skywalking/blob/master/.github/workflows/skywalking.yaml))


javeme:
@imbajin Is this PR useless? if so we can close it

imbajin:
> @imbajin Is this PR useless? if so we can close it

depend on shall we add `required` in ci..

.asf.yaml
```diff
@@ -0,0 +1,51 @@
...
+    rebase: true
+  protected_branches:
+    master:
+      required_status_checks:
+        # strict means "Require branches to be up to date before merging".
+        strict: true
+        # contexts are the names of checks that must pass
+        contexts:
+          - hugegraph ci/build
+          - hugegraph-ci/build
```
javeme, line 43: TODO: make some CI required

javeme (commented):
In view of the fact that it is currently impossible to enable required context since it may skip some components CI due to the trigger mechanism, we just add an issue to track it and close this PR?

imbajin:
both are fine (sine it have no diff now)

.asf.yaml
imbajin: enable it(the rule) after the package rename done

Linary (dismissed):

imbajin:
@JackyYangPassion review the config with license checker, I found some files are excluded:
image

wonder to know if this is generally allowed to be excluded by the Apache projects/community?

seems we should exclude them carefully~

javeme (dismissed):

Linary (approved):

imbajin:
cherry pick the check in #428 for hubble

javeme (approved):

PR #326, opened 2022-08-17
fix: Line fetch some value null

BestBurning:
fix: Line fetch some value null

### DIFF
--- a/hugegraph-loader/src/main/java/com/baidu/hugegraph/loader/reader/file/OrcFileLineFetcher.java
+++ b/hugegraph-loader/src/main/java/com/baidu/hugegraph/loader/reader/file/OrcFileLineFetcher.java
@@ -123,11 +123,18 @@ public class OrcFileLineFetcher extends FileLineFetcher {
         if (!this.recordReader.hasNext()) {
             return null;
         }
-
+        Object[] values = null;
         this.row = this.recordReader.next(this.row);
-        Object[] values = this.inspector.getStructFieldsDataAsList(this.row)
-                                        .stream().map(Object::toString)
-                                        .toArray();
+        try {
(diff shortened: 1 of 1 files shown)


hugegraph-loader/src/main/java/com/baidu/hugegraph/loader/reader/file/OrcFileLineFetcher.java
```diff
@@ -123,11 +123,18 @@ public Line fetch() throws IOException {
...
-
+        Object[] values = null;
         this.row = this.recordReader.next(this.row);
-        Object[] values = this.inspector.getStructFieldsDataAsList(this.row)
-                                        .stream().map(Object::toString)
-                                        .toArray();
+        try {
+            values = this.inspector.getStructFieldsDataAsList(this.row)
+                    .stream()
+                    .map(o -> o != null ? o.toString() : "")
```
coderzc, line 131: Filter null value？

hugegraph-loader/src/main/java/com/baidu/hugegraph/loader/reader/file/OrcFileLineFetcher.java
```diff
@@ -123,11 +123,18 @@ public Line fetch() throws IOException {
...
+                    .toArray();
+        } catch (NullPointerException e) {
```
javeme, line 133: I think it won't throw NullPointerException if we translated null to "", right?

javeme (commented):
Thank you for your contribution

javeme:
Looks the same as PR https://github.com/apache/incubator-hugegraph-toolchain/pull/321

imbajin:
close it since no update for long time

PR #335, opened 2022-10-08
Add maven-clean-plugin for hubble

simon824:
Avoid creating too many temporary files by `cp` and add `maven-clean-plugin` for hubble to clean package files.


imbajin (approved):
thanks, shall we put the plugin in root's pom?

simon824:
> thanks, shall we put the plugin in root's pom?

but seems different subprojects have different files to clean, maybe it's better to separate?

javeme (approved):

PR #339, opened 2022-10-20
chore: update some params to fix CI error

imbajin:
now we have some problems to solve: (TODO)
1. we need adopt/update the config path for server(master) first (due to we support multi graph in 0.13, and the dir tree changed)
2. we need update the client's APIVersion to meet the check (like CI in loader)
3. last, after the server changed version to apache 1.0, we should also adapt it (later)

PS: could we use one commit ID to unify the CI? or use a action template

### DIFF
--- a/hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
+++ b/hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
@@ -12,7 +12,7 @@ HUGEGRAPH_GIT_URL="https://github.com/apache/hugegraph.git"
 git clone --depth 100 ${HUGEGRAPH_GIT_URL} hugegraph
 cd hugegraph
 git checkout "${COMMIT_ID}"
-mvn package -DskipTests -Dmaven.javadoc.skip=true | grep -v "Downloading\|Downloaded\|Progress"
+mvn package -DskipTests -Dmaven.javadoc.skip=true -ntp
 # TODO: lack incubator after apache package release (update it later)
 mv hugegraph-*.tar.gz ../
 cd ../
--- a/hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
+++ b/hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
@@ -3,24 +3,25 @@
 set -ev
 
 if [[ $# -ne 1 ]]; then
-    echo "Must pass commit id of hugegraph repo"
-    exit 1
+    echo "Must input exist git commit id of hugegraph server" && exit 1
 fi
 
 COMMIT_ID=$1
-HUGEGRAPH_GIT_URL="https://github.com/hugegraph/hugegraph.git"
+HUGEGRAPH_GIT_URL="https://github.com/apache/hugegraph.git"
 
 git clone --depth 100 ${HUGEGRAPH_GIT_URL} hugegraph
 cd hugegraph
 git checkout "${COMMIT_ID}"
-mvn package -DskipTests
+mvn package -DskipTests -Dmaven.javadoc.skip=true | grep -v "Downloading\|Downloaded\|Progress"
+# TODO: lack incubator after apache package release (update it later)
 mv hugegraph-*.tar.gz ../
 cd ../
 rm -rf hugegraph
-tar -zxvf hugegraph-*.tar.gz
+tar zxvf hugegraph-*.tar.gz
 
 HTTPS_SERVER_DIR="hugegraph_https"
 mkdir ${HTTPS_SERVER_DIR}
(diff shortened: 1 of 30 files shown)


z7658329:
> now we have some problems to solve: (TODO)
> 
> 1. we need adopt/update the config path for server(master) first (due to we support multi graph in 0.13, and the dir tree changed)
> 2. we need update the client's APIVersion to meet the check (like CI in loader)
> 3. last, after the server changed version to apache 1.0, we should also adapt it (later)
> 
> PS: could we use one commit ID to unify the CI? or use a action template

i can try to slove:
2. we need update the client's APIVersion to meet the check (like CI in loader)
3. last, after the server changed version to apache 1.0, we should also adapt it (later)

is it ok?

imbajin:
> i can try to slove: 2. we need update the client's APIVersion to meet the check (like CI in loader) 3. last, after the server changed version to apache 1.0, we should also adapt it (later)
> 
> is it ok?

fine, but step3 need wait for **server done**, the 1,2 step are now blocking the merge process (The first one is not troublesome. You can try it 😃 )

hugegraph-loader/assembly/travis/install-hugegraph-from-tar.sh
```diff
@@ -3,18 +3,19 @@
...
+    echo "Must input exist git commit id of hugegraph server" && exit 1
 fi
 
 VERSION=$1
 HUGEGRAPH_LINK="https://github.com/hugegraph/hugegraph/releases/download/v${VERSION}/hugegraph-${VERSION}.tar.gz"
 
 wget "${HUGEGRAPH_LINK}" || exit 1
-tar -zxvf hugegraph-${VERSION}.tar.gz
+# TODO: lack incubator after apache package release (update it later)
+tar zxvf hugegraph-${VERSION}.tar.gz
```
javeme, line 14: we can define an env var for it like PACKAGE_PREFIX="incubator-hugegraph"

hugegraph-loader/assembly/travis/install-hugegraph-from-tar.sh
```diff
@@ -3,18 +3,19 @@
 set -ev
 
 if [[ $# -ne 1 ]]; then
-    echo "Must pass server version of hugegraph"
-    exit 1
+    echo "Must input exist git commit id of hugegraph server" && exit 1
```
javeme, line 6: Must input an existing commit id of hugegraph server

hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
```diff
@@ -3,24 +3,25 @@
...
 cd hugegraph
 git checkout "${COMMIT_ID}"
-mvn package -DskipTests
+mvn package -DskipTests -Dmaven.javadoc.skip=true | grep -v "Downloading\|Downloaded\|Progress"
+# TODO: lack incubator after apache package release (update it later)
 mv hugegraph-*.tar.gz ../
 cd ../
 rm -rf hugegraph
-tar -zxvf hugegraph-*.tar.gz
+tar zxvf hugegraph-*.tar.gz
```
javeme, line 20: update "hugegraph-*" like the script in hugegraph server

hugegraph-loader/assembly/travis/install-hugegraph-from-tar.sh
imbajin: @z7658329 you could also do it in the next PR

imbajin:
@simon824 seems the hubble CI failed due to some reason, could u take a look with it?

image

simon824:
@imbajin  It seems that the npm source address cannot be found, it may happen by accident,  have you tried rerunning ci?

imbajin:
> @imbajin  It seems that the npm source address cannot be found, it may happen by accident,  have you tried rerunning ci?
> 
> 

sure，try rerun it 3 times，and the other PR didnt have this problem？（cache？ skip the real build？）

z7658329:
> 
@imbajin @simon824 
it seems works, see #351 
the reason i think is maven-antrun-plugin cant  resolve project.parent.basedir
4acc403944d23e44a38d71d53c365e7

imbajin:
After all ci pass，merge it first，then handle the ignore test in。separate PR

hugegraph-loader/assembly/travis/install-hugegraph-from-tar.sh
z7658329: > incubator-hugegraph
like this?
image

javeme (approved):

simon824 (approved):

zyxxoo (approved):

PR #349, opened 2022-10-30
change limit type long to int

zyxxoo:


imbajin (commented):
seems I try to change it in server formerly, but suspend for some reason....  @javeme

imbajin (approved):

javeme (approved):

PR #353, opened 2022-11-02
refact: adopt & use apache version in toolchain

z7658329:
change package to org.apache.hugegraph

PR #356, opened 2022-11-04
chore: enable flatten while building

imbajin:
this plugin make the `$ {revision}` clean & work well, otherwise the maven will treat the variable as a string:

image


coderzc (dismissed):

javeme (dismissed):

.github/workflows/loader-ci.yml
```diff
@@ -62,8 +62,7 @@ jobs:
 
       - name: Run test
         run: |
-          echo "##### Ignore Client Test Now ######"
-          mvn test -P unit,file,hdfs,jdbc -pl hugegraph-client,hugegraph-loader -fn
+          mvn test -P unit,file,hdfs,jdbc
```
imbajin, line 65: @z7658329 copy like this in #353   (`-fn` will ignore the mvn compile error)

javeme (approved):

coderzc (approved):

PR #359, opened 2022-11-08
chore: remove oracle dependency

simon824:
Remove oracle dependency


imbajin:
without the `ojdbc`, the diff/influence is?

simon824:
> without the `ojdbc`, the diff/influence is? (why we need it before, used in test?)

It seems that there is no dependency on ojdbc in ut, just for including ojdbc when packaging.

imbajin:
> > without the `ojdbc`, the diff/influence is? (why we need it before, used in test?)
> 
> It seems that there is no dependency on ojdbc in ut, maybe just for including ojdbc when packaging?

seems a little strange, @javeme  (is it useful? otherwise remove it now)

javeme (approved):

PR #363, opened 2022-11-10
chore: update spark&flink scope to provided

simon824:
spark&flink loader needs to depend on the local spark&flink environment, so it's not necessary to introduce these dependencies in toolchains.


imbajin (approved):

imbajin:
find another problem, currently the package file is too big (seems we should config maven-assembly-plugin)

image

PR #358, opened 2022-11-07
feat(hbase):  support gen HFile for hbase v2 (BETA)

haohao0103:
close #357 

1、Support write vertex/edge directly to KV storage
2、Just support customString and customNumber ID now
3、submit the loader code for bypass server for hbase writing

PR #364, opened 2022-11-10
adapt to org.apache package

z7658329:


z7658329:
seems config COMMITID fall behind  server's latest commitid

imbajin:
> seems config COMMITID fall behind server's latest commitid

could update it, however how could we know it's the root cause? Seems other PR use the same commit but passed

we should better let server print the error info when it failed to start

z7658329:
> > seems config COMMITID fall behind server's latest commitid
> 
> could update it, however how could we know it's the root cause? Seems other PR use the same commit but passed
> 
> we should better print the server.log error when it failed to start

yes ,we can test in  local and fix this problem

z7658329:
looks strange, when start HugeGraphServer with https protocol,the loader ci failed while the tools ci success:

loader-ci:
https://github.com/apache/incubator-hugegraph-toolchain/actions/runs/3434429029/jobs/5725726685

tools-ci:
https://github.com/apache/incubator-hugegraph-toolchain/actions/runs/3434429034/jobs/5725726848

image

imbajin:
> looks strange, when start HugeGraphServer with https protocol,the loader ci failed while the tools ci success:

loader ci fixed in #358, just ignore it now

z7658329:
> 



> > looks strange, when start HugeGraphServer with https protocol,the loader ci failed while the tools ci success:
> 
> loader ci fixed in #358, just ignore it now

get it, should i copy the pom.xml to this branch?

z7658329:
i have try to start hugegraph server with https protocal in local ,success, seems no matter with pr #358 
image

PR #387, opened 2022-12-03
rename binary

z7658329:


z7658329:
image

image

javeme (approved):
LGTM

imbajin (approved):

hugegraph-dist/pom.xml
```diff
@@ -14,7 +14,7 @@
 
     <properties>
         <release.name>${project.parent.artifactId}</release.name>
-        <final.name>${release.name}-${project.version}</final.name>
+        <final.name>apache-${release.name}-incubating-${project.version}</final.name>
```
imbajin, line 17: @z7658329 should use one param for these modules?

seems we just define it in root-pom is OK?

pom.xml
```diff
@@ -61,7 +61,7 @@
     <properties>
         <revision>1.0.0</revision>
         <release.name>${project.artifactId}</release.name>
-        <final.name>${release.name}-${project.version}</final.name>
+        <final.name>apache-${release.name}-incubating-${project.version}</final.name>
```
imbajin, line 64: the child module will extend it automatically

PR #390, opened 2022-12-05
chore: use param in antrun to support build in windows

z7658329:


hugegraph-loader/pom.xml
```diff
@@ -648,19 +648,12 @@
...
-                                </echo>
-                                <exec executable="${shell-executable}"
-                                      dir="${project.basedir}" failonerror="true">
-                                    <arg line="./dist.sh"/>
-                                </exec>
-                            </tasks>
+                            <target>
+                                <tar destfile="${project.basedir}/${final.name}.tar.gz"
+                                     basedir="${project.basedir}"
+                                     includes="${final.name}/**"
```
imbajin, line 654: old name is `hugegraph-loader tar.gz`

and the new pkg name is? `final.name`?

z7658329: > 

 i have fix all tools binary rename in pr #387

javeme (approved):

imbajin (approved):

PR #394, opened 2022-12-08
fix(dist): bin package lack execute permission

z7658329:


hugegraph-loader/pom.xml
```diff
@@ -650,9 +650,11 @@
                         <configuration>
                             <target>
                                 <tar destfile="${project.basedir}/${final.name}.tar.gz"
```
coderzc, line 652:
```suggestion
                                <tar destfile="${project.basedir}/target/${final.name}.tar.gz"
```

hugegraph-tools/pom.xml
```diff
@@ -235,9 +235,11 @@
                         <configuration>
                             <target>
                                 <tar destfile="${project.basedir}/${final.name}.tar.gz"
```
coderzc, line 237:
```suggestion
                                <tar destfile="${project.basedir}/target/${final.name}.tar.gz"
```

hugegraph-loader/pom.xml
```diff
@@ -650,9 +650,11 @@
...
-                                     basedir="${project.basedir}"
-                                     includes="${final.name}/**"
-                                     compression="gzip"/>
+                                     compression="gzip">
+                                    <tarfileset dir="${project.basedir}" filemode="755">
```
coderzc, line 654: why is `${project.basedir}`?

z7658329: because plugins will generate sources in this path

hugegraph-loader/pom.xml
z7658329: the generate binary's dest path should not be project's target dir ? @coderzc @imbajin

imbajin: > the generate binary's dest path should not be project's target dir ? @coderzc @imbajin

it should under the target, but be careful to check the relative influence like in computer's CI or some script

coderzc (approved):

hugegraph-tools/pom.xml
```diff
@@ -187,11 +187,6 @@
     </dependencies>
 
     <build>
-        <directory>output</directory>
-        <sourceDirectory>src/main/java</sourceDirectory>
-        <!--<testSourceDirectory>src/test/java</testSourceDirectory>-->
-        <outputDirectory>output/classes</outputDirectory>
-        <!--<testOutputDirectory>output/test-classes</testOutputDirectory>-->
```
imbajin, line 194: @javeme ensure is it useless now?

imbajin (approved):

PR #402, opened 2022-12-25
fix(loader): wrong shell symbol in spark script

JackyYangPassion:
fix spark loader sh error


imbajin (approved):
thanks，how could we find these bugs~

simon824 (approved):

javeme (approved):

PR #405, opened 2023-01-03
chore: add NOTICE file

simon824:


javeme (approved):

hugegraph-loader/assembly/static/bin/hugegraph-spark-loader.sh
```diff
@@ -1,4 +1,4 @@
-)`:
+#!/bin/bash
```
imbajin, line 1: nice catch:)

NOTICE
```diff
@@ -0,0 +1,5 @@
+Apache HugeGraph(incubating)
+Copyright 2022-2023 The Apache Software Foundation
```
imbajin, line 2: ensure the start year is right?

imbajin (approved):
thanks, the computer module also need it?

PR #408, opened 2023-01-04
support jdk11,remove config java.ext.dirs

z7658329:


z7658329:
i have test, success in jdk8/jdk11

imbajin (approved):
thanks, could also check the hubble and loader

PR #434, opened 2023-03-01
refact: clean extra store file in all modules

imbajin:
TODO: 
- [ ] we need use a better way to download it when building binary package
- [ ] ensure the **deleted** file `keystore` or `license.json` is no longer needed @javeme 

@simon824 download the file in building or let user download it manually when they need `https`? (and we add a doc/link for it)


hugegraph-client/src/main/java/org/apache/hugegraph/util/CommonUtil.java
```diff
@@ -41,4 +46,14 @@ public static void checkMapClass(Object object, Class<?> kClass,
...
         }
     }
+
+    public static void downloadFileByUrl(String url, String destPath) {
+        int connectTimeout = 5000;
+        int readTimeout = 5000;
+        try {
+            FileUtils.copyURLToFile(new URL(url), new File(destPath), connectTimeout, readTimeout);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
javeme, line 56: add some message to translate the exception

hugegraph-loader/src/test/java/org/apache/hugegraph/loader/test/functional/FileLoadTest.java
```diff
@@ -2426,6 +2427,10 @@ public void testParquetCompressFile() {
         };
         if (this.ioUtil instanceof HDFSUtil) {
             HDFSUtil hdfsUtil = (HDFSUtil) this.ioUtil;
+            String url = "https://github.com/apache/incubator-hugegraph-doc/" +
```
javeme, line 2430: define a const var for the base url, may be it may be changed somehow, so that we don't need update everywhere

javeme (dismissed):

imbajin:
image

zyxxoo (approved):

javeme (approved):

PR #440, opened 2023-03-15
chore(hubble): use latest code in Dockerfile

simon824:
TODO:

- [ ] update doc in website with server (doc need a big update)

### DIFF
--- a/hugegraph-hubble/Dockerfile
+++ b/hugegraph-hubble/Dockerfile
@@ -5,7 +5,9 @@
 # The ASF licenses this file to You under the Apache License, Version 2.0
 # (the "License"); you may not use this file except in compliance with
 # the License.  You may obtain a copy of the License at
-#     http://www.apache.org/licenses/LICENSE-2.0
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


hugegraph-hubble/Dockerfile
```diff
@@ -5,30 +5,47 @@
...
+    && cd /pkg/hugegraph-hubble/ \
+    && mvn package -e -B -ntp -DskipTests -Dmaven.javadoc.skip=true
+    
+
+FROM openjdk:11-slim
 
+RUN set -x \
+    && apt-get -q update \
+    && apt-get -q install -y --no-install-recommends --no-install-suggests curl lsof g++ gcc \
+    && apt-get clean
```
imbajin, line 39: maybe we don't need `curl / lsof / g++ / gcc` here? 

maybe we could remove the whole step

hugegraph-hubble/Dockerfile
```diff
@@ -5,30 +5,47 @@
...
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
 
-FROM ubuntu:xenial
-
+FROM maven:3.9.0-eclipse-temurin-11 AS build
+
```
imbajin, line 19: extra spaces here?

hugegraph-hubble/Dockerfile
```diff
@@ -5,30 +5,47 @@
...
+RUN set -x \
+    && apt-get -q update \
+    && apt-get -q install -y --no-install-recommends --no-install-suggests curl lsof g++ gcc \
+    && apt-get clean 
+    
+COPY --from=build /pkg/hugegraph-hubble/apache-hugegraph-hubble-incubating-*/ /hubble
+
+RUN set -e \
+    && sed -i 82{s/nohup//g} /hubble/bin/start-hubble.sh \
+    && sed -i 83{s/\>.*//g} /hubble/bin/start-hubble.sh \
```
imbajin, line 45: 1. why we need to remove `nohup`, seems it's enough just remove `&` to avoid it running in bg?
2. consider add a option like server, use `start-hubble.sh -d false` is better? (**latest code** line may change easily)

javeme (approved):

imbajin (commented):
We also add the short usage info in `README` (Both in root & hubble) 

And I'll test it soon

imbajin (approved):

imbajin (approved):
could upload to `hugegraph/hubble:latest` now

PR #447, opened 2023-03-29
set default data  when create graph

wanganjuan:
set default data  when create graph
image


hugegraph-hubble/hubble-fe/src/stores/GraphManagementStore/graphManagementStore.ts
```diff
@@ -220,6 +220,14 @@ export class GraphManagementStore {
     this.editGraphData.password = this.graphData[index].password;
   }
 
+  @action
+  fillInGraphDataDefaultConfig() {
+    this.newGraphData.name = 'test';
+    this.newGraphData.graph = 'hugegraph';
+    this.newGraphData.host = '0.0.0.0';
+    this.newGraphData.port = '18080';
```
imbajin, line 228: use `8080` for default

wanganjuan: > use `8080` for default

fixed

simon824 (approved):

Issue #456, opened 2023-05-08
[Enhancement] keep hadoop-hdfs-client and hadoop-common version consistent


simon824:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-hugegraph-toolchain/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Feature Description (功能描述)

refer to  #454 
hadoop-common 3.3.1
hadoop-hdfs-client 2.10.1

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

imbajin:
maybe unify them with one version like `${hdfs.version}`

chenlixuan:
I can help to improve it.

PR #457, opened 2023-05-10
chore: keep hadoop-hdfs-client and hadoop-common version consistent

chenlixuan:
refer to #456

### DIFF
--- a/hugegraph-loader/pom.xml
+++ b/hugegraph-loader/pom.xml
@@ -40,6 +40,7 @@
         <release.name>${project.artifactId}</release.name>
         <final.name>apache-${release.name}-incubating-${project.version}</final.name>
         <jackson.version>2.12.3</jackson.version>
+        <hdfs.version>3.3.1</hdfs.version>
         <assembly.dir>${project.basedir}/assembly</assembly.dir>
         <assembly.descriptor.dir>${assembly.dir}/descriptor</assembly.descriptor.dir>
         <assembly.static.dir>${assembly.dir}/static</assembly.static.dir>
@@ -307,7 +308,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>3.3.1</version>
+            <version>${hdfs.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.slf4j</groupId>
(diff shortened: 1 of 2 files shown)


hugegraph-loader/pom.xml
```diff
@@ -307,7 +308,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>3.3.1</version>
+            <version>${hdfs.version}</version>
```
simon824, line 311: `hadoop.version` maybe better

simon824 (approved):

hugegraph-loader/pom.xml
```diff
@@ -40,6 +40,7 @@
         <release.name>${project.artifactId}</release.name>
         <final.name>apache-${release.name}-incubating-${project.version}</final.name>
         <jackson.version>2.12.3</jackson.version>
+        <hadoop.version>3.3.1</hadoop.version>
```
imbajin, line 43: Could check whether there are other similar Hadoop dependencies in the project at the same time. If there are multiple module dependencies, you can put the version attribute under the root (pom)

imbajin (approved):

imbajin (changes requested):
Need update the dependency file to fix the ci error @simon824 we should add a **PR template** and tell the devs how to modify it 

image

imbajin (changes requested):
Need update the dependency file to fix the ci error @simon824 we should add a **PR template** and tell the devs how to modify it 

image

simon824:
hi @chenlixuan , in order to check the LICENSE conveniently, when we update the dependency, we need also update this file  at the same time (https://github.com/chenlixuan/incubator-hugegraph-toolchain/blob/master/hugegraph-dist/scripts/dependency/known-dependencies.txt #L92) and make the version of dependencies in the file consistent with pom.xml

hugegraph-loader/pom.xml
chenlixuan: I will do it later.

simon824 (approved):

hugegraph-loader/pom.xml
simon824: > I will do it later.

@chenlixuan We can do it in the next pr,  unify the dependencies versions of the entire project.

imbajin (approved):

PR #510, opened 2023-08-19
fix(loader): resolve error loader script

### LINKED ISSUE #509: [Bug]  Exception Occurs When Running hugegraph-loader.sh Script from the HugeGraph official tutorial.
### Bug Type (问题类型)

exception / error (异常报错)

### Before submit

- [X] I had searched in the [issues](https://github.com/apache/hugegraph-toolchain/issues) and found no similar issues.

### Environment (环境信息)

- Server Version: v1.0.0
- Toolchain Version: v1.0.0
- Data Size: 8 vertices, 6 edges
- Operating System: Ubuntu 22.04
- Shell Interpreter: sh (also tried bash)

### Expected & Actual behavior (期望与实际表现)

Issue Description: I'm encountering an unexpected behavior when attempting to run the hugegraph-loader.sh script from the HugeGraph official tutorial. The issue seems to be specific to different shell environments.
Command Example:
`$ sh bin/hugegraph-loader.sh -g hugegraph -f example/file/struct.json -s example/file/schema.groovy`
When executing the above command using the sh shell, I encounter the following exceptions:
```
hugegraph-loader.sh: 19: Bad substitution
hugegraph-loader.sh: 37: set: Illegal option -o pipefail

```
Attempted Solutions:

The command runs successfully using the bash shell.

_No response_

_No response_

lzyxx77:
I have read the CLA Document and I hereby sign the CLA
#509
Changes Proposed:

I've made modifications to the hugegraph-loader.sh script to ensure compatibility with various shell interpreters, including bash, dash, and ash. issue#509  was related to the fact that the /bin/sh symlink on many Unix-like systems can point to different shell interpreters such as bash, dash, or ash. This can cause problems when scripts rely on specific syntax features that are not universally supported across different shells.

The specific changes made include:

Adjusting syntax that might not be supported in dash to be compatible.
Ensuring usage of POSIX shell-compatible syntax throughout the script.

Testing:

I've tested the modified hugegraph-loader.sh script in various shell environments including bash, dash, and ash, and it works as intended without any syntax errors or unexpected behavior.

Examples:

I'm encountering an unexpected behavior when attempting to run the hugegraph-loader.sh script from the HugeGraph official tutorial. The issue seems to be specific to different shell environments.
use shl:

image

use bash:
image

So I  use the old style to ensure sh works as intended without any syntax errors or unexpected behavior.

image

The reson why I use the old style to ensure is follow：

In Ubuntu, /bin/sh is by default symlinked to dash, which is a different shell than bash. This change was made to improve script execution speed, as bash is more complex. dash was ported from NetBSD to Linux and recommended for use as /bin/dash for faster script execution. dash is faster than bash but has fewer features, aligning more closely with the POSIX standard. Specifically, in this context, the [[ construct is an extension of bash used for more advanced conditional testing, and dash does not support it. Therefore, when running scripts in the sh environment, it results in the [[: not found error.

Of course, you can also change the symlink for sh in your Ubuntu system to point to bash so that you don't need to modify sh scripts.


hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
```diff
@@ -16,11 +16,11 @@
...
 #
 abs_path() {
-    SOURCE="${BASH_SOURCE[0]}"
-    while [[ -h "$SOURCE" ]]; do
+    SOURCE="$0"
+    while [ -h "$SOURCE" ]; do
         DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
         SOURCE="$(readlink "$SOURCE")"
-        [[ ${SOURCE} != /* ]] && SOURCE="$DIR/$SOURCE"
+        [ "${SOURCE}" != /* ] && SOURCE="$DIR/$SOURCE"
```
imbajin, line 23: why we use the old style？

show some example thx~

hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
```diff
@@ -78,4 +81,4 @@ export JVM_OPTS="$JVM_OPTS -Xmx10g -cp $LOADER_CLASSPATH"
 
 exec ${JAVA} -Dname="HugeGraphLoader" -Dloader.home.path=${TOP} -Dlog4j.configurationFile=${CONF}/log4j2.xml \
 -Djava.library.path=${NATIVE} \
-${JVM_OPTS} org.apache.hugegraph.loader.HugeGraphLoader ${VARS}
+${JVM_OPTS} org.apache.hugegraph.loader.HugeGraphLoader ${VARS}
```
imbajin, line 84: lack empty line （in the end）

hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
lzyxx77: Because my Ubuntu system's 'sh' command is not linked to 'bash', using the 'sh' command directly doesn't work. I believe others might encounter the same issue, so I've used the old style. If this problem isn't significant, I will close this pull request~

hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
lzyxx77: thanks，I willl add it！

hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
javeme: maybe it just caused by `"${BASH_SOURCE[0]}"`?

javeme:
Thanks for your contribution

hugegraph-loader/assembly/static/bin/hugegraph-loader.sh
lzyxx77: > maybe it just caused by `"${BASH_SOURCE[0]}"`?

thinks for your reply！However, when I only modify "${BASH_SOURCE[0]}" and then execute it with sh, it still fails. The reason in follow:

In Ubuntu, /bin/sh is by default symlinked to dash, which is a different shell than bash. This change was made to improve script execution speed, as bash is more complex. dash was ported from NetBSD to Linux and recommended for use as /bin/dash for faster script execution. dash is faster than bash but has fewer features, aligning more closely with the POSIX standard. Specifically, in this context, the [[ construct is an extension of bash used for more advanced conditional testing, and dash does not support it. Therefore, when running scripts in the sh environment, it results in the [[: not found error.

Of course, you can also change the symlink for sh in your Ubuntu system to point to bash so that you don't need to modify sh scripts.

imbajin (approved):

PR #539, opened 2023-11-30
feat: support edgeexistence api

msgui:
follow https://github.com/apache/incubator-hugegraph/pull/2242


hugegraph-client/src/main/java/org/apache/hugegraph/driver/TraverserManager.java
```diff
@@ -538,4 +541,10 @@ public Iterator<Edge> iteratorEdges(Shard shard, int sizePerPage) {
             return this.edges(shard, page, sizePerPage);
         });
     }
+
+    public List<Edge> edgeExistence(Object sourceId, Object targetId, String edgeLabel,
```
javeme, line 545: also add this 2 versions:
```java
edgeExistence(Object sourceId, Object targetId)
edgeExistence(Object sourceId, Object targetId, String edgeLabel)
```

PR #590, opened 2024-03-13
feat(client): support user defined OKHTTPClient configs

zhenyuT:
Follow https://github.com/apache/incubator-hugegraph-commons/pull/140, add a callback builder for user defined configs in (OK)HTTPClient

Before:
(Users)We have no way to modify the `(HugeClient(RestClient(HttPClient)))` configs, we have to modify `hugegraph-common` which is really inflexible & hard to use...

After:
```java
            HugeClient.builder(url, graph)
                               .configUser(username, password)
                               .configTimeout(timeout)
                               .configHttpBuilder(x -> x.changeHTTPClientConfigs(....))
                               .build();
```


pom.xml
```diff
@@ -99,7 +99,7 @@
 
     <properties>
         <revision>1.2.0</revision>
-        <hugegraph.common.version>1.2.0</hugegraph.common.version>
+        <hugegraph.common.version>1.3.0</hugegraph.common.version>
```
imbajin, line 102: @JackyYangPassion we need update the `rivision` in root before release

liuxiaocs7 (approved):
+1

imbajin (approved):

PR #385, opened 2022-11-28
refact: fix some code style with check plugin

seagle-yuan:
fix checkstyle check

PR #601, opened 2024-04-14
chore: followup checkstyle check

liuxiaocs7:
## Purpose of the PR

- follow up and close #385 

## Verifying these changes

- [x] Trivial rework / code cleanup without any test coverage. (No Need)
    - xxx

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-client/src/main/java/org/apache/hugegraph/serializer/direct/HBaseSerializer.java
+++ b/hugegraph-client/src/main/java/org/apache/hugegraph/serializer/direct/HBaseSerializer.java
@@ -49,12 +49,12 @@ public class HBaseSerializer {
 
     public byte[] getKeyBytes(GraphElement e) {
         byte[] array = null;
-        if (e.type() == "vertex" && e.id() != null) {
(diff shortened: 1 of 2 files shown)


hugegraph-client/src/main/java/org/apache/hugegraph/serializer/direct/HBaseSerializer.java
```diff
@@ -49,12 +49,12 @@ public HBaseSerializer(HugeClient client, int vertexPartitions, int edgePartitio
 
     public byte[] getKeyBytes(GraphElement e) {
         byte[] array = null;
-        if (e.type() == "vertex" && e.id() != null) {
+        if ("vertex".equals(e.type()) && e.id() != null) {
```
javeme, line 52: can we define a const var for it?

Issue #641, opened 2024-11-15
[Format] Align `.editorconf` file to server


Thespica:
### Search before asking

- [X] I had searched in the [feature](https://github.com/apache/incubator-hugegraph-toolchain/issues?q=is%3Aissue+label%3A%22Feature%22) and found no similar feature requirement.

### Feature Description (功能描述)

`.editorconf` [file of toolchain](https://github.com/apache/incubator-hugegraph-toolchain/blob/master/.editorconfig) is out date of [the same file in server](https://github.com/apache/incubator-hugegraph-toolchain/blob/master/.editorconfig).

Need a PR to update this file, and up date related code format.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

imbajin:
fixed by #648

PR #648, opened 2025-01-03
feat(loader): support jdbc date type & sync .editorconfig

MuLeiSY2021:
## Purpose of the PR
add jdbc date type support include DATE、TIME、 DATETIME、 TIMESTAMP and YEAR

- fix [#298](https://github.com/apache/incubator-hugegraph-toolchain/issues/298 

- Add a test ,which show all different type import by jdbc
## Main Changes

- I change the logic of if value is date,KafkaSource can't instaceof FileSource,So I turned them into juxtaposition logic,and add JDBC source convert
- In mysql cases,there are many type of date,and if just type casting，the error gonna happend,so I need new var to solve this problem
## Verifying these changes

- [x] Trivial rework / code cleanup without any test coverage. (No Need)
    - testJdbcSqlDateConvert

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,17 +34,17 @@ jobs:
       fail-fast: false
       matrix:
         language: [ 'java', 'javascript', 'python' ]
+        JAVA_VERSION: [ '11' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v3
-
+    - name: Fetch Code
+      uses: actions/checkout@v4
     - name: Setup Java JDK
       uses: actions/setup-java@v3
       with:
+        java-version: ${{ matrix.JAVA_VERSION }}
         distribution: 'zulu'
-        java-version: '8'
 
     - name: use staged maven repo settings
       if: ${{ env.USE_STAGE == 'true' }}
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -34,24 +34,24 @@ jobs:
       fail-fast: false
       matrix:
         language: [ 'java', 'javascript', 'python' ]
-        JAVA_VERSION: [ '11' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
 
     steps:
-    - name: Fetch Code
-      uses: actions/checkout@v4
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
     - name: Setup Java JDK
       uses: actions/setup-java@v3
       with:
-        java-version: '8'
         distribution: 'zulu'
+        java-version: '8'
 
     - name: use staged maven repo settings
       if: ${{ env.USE_STAGE == 'true' }}
       run: |
         cp $HOME/.m2/settings.xml /tmp/settings.xml
         mv -vf .github/configs/settings.xml $HOME/.m2/settings.xml
-        
+
     - name: Use Node.js 16
       uses: actions/setup-node@v3
       with:
@@ -64,11 +64,11 @@ jobs:
         # If you wish to specify custom queries, you can do so here or in a config file.
         # By default, queries listed here will override any specified in a config file.
         # Prefix the list here with "+" to use these queries and those in the config file.
-        
+
         # Details on CodeQL's query packs refer to : https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-queries-in-ql-packs
         # queries: security-extended,security-and-quality
 
-        
+
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - if: matrix.language == 'python' || matrix.language == 'javascript'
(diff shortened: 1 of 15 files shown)


imbajin:
Thanks a lot for your contribution, could refer the doc to format the code:
- https://github.com/apache/hugegraph/wiki/HugeGraph-Code-Style-Guide
- https://github.com/apache/hugegraph/wiki/The-style-config-for-HugeGraph-in-IDEA#%E7%AC%AC%E9%9B%B6%E6%AD%A5-github-prgit-%E6%B5%81%E7%A8%8B%E4%BD%BF%E7%94%A8

MuLeiSY2021:
@imbajin 
> Thanks a lot for your contribution, could refer the doc to format the code: (to avoid a lot of unexpected changes)

Thank you for the reminder. After I checked my code style, I'm pretty sure I have already opened that configuration.  
I think there might be a bug:  
When I pasted the `.editorconfig` from the "apache/incubator-hugegraph" repository, it worked as expected, just like the original file.  
However, the `.editorconfig` from the "apache/incubator-hugegraph-toolchain" repository didn't work.  
Maybe we should migrate the `.editorconfig` from "apache/incubator-hugegraph" to "apache/incubator-hugegraph-toolchain" to solve this problem.

Also, I noticed that:  
```
# Ignore the IDEA unsupported warning & it works well (indeed) 
continuation_indent_size = 16
```
It doesn't work on my IDEA version: "IntelliJ IDEA 2024.3.1 (Ultimate Edition)".  
After referring to this article
> [https://youtrack.jetbrains.com/issue/IJPL-148288/Continuation-indent-size-from-IDE-setting-is-not-used-when-using-.editorconfig](https://youtrack.jetbrains.com/issue/IJPL-148288/Continuation-indent-size-from-IDE-setting-is-not-used-when-using-.editorconfig)
@thorsten.schoening:
Thanks for reporting this issue. It seems we should use the continuations setting from the IDE's code style setting for your case.
As a workaround, you could add ij_continuation_indent_size = 8 in .editorconfig. to fix this issue.

I changed "continuation_indent_size" to "ij_continuation_indent_size", and it worked well,maybe i should ask another issue

imbajin:
> @imbajin
> 
> > Thanks a lot for your contribution, could refer the doc to format the code: (to avoid a lot of unexpected changes)
> 
> Thank you for the reminder. After I checked my code style, I'm pretty sure I have already opened that configuration. I think there might be a bug: When I pasted the `.editorconfig` from the "apache/incubator-hugegraph" repository, it worked as expected, just like the original file. However, the `.editorconfig` from the "apache/incubator-hugegraph-toolchain" repository didn't work. Maybe we should migrate the `.editorconfig` from "apache/incubator-hugegraph" to "apache/incubator-hugegraph-toolchain" to solve this problem.
> 
> Also, I noticed that:
> 
> ```
> # Ignore the IDEA unsupported warning & it works well (indeed) 
> continuation_indent_size = 16
> ```
> 
> It doesn't work on my IDEA version: "IntelliJ IDEA 2024.3.1 (Ultimate Edition)". After referring to this article
> 
> > [youtrack.jetbrains.com/issue/IJPL-148288/Continuation-indent-size-from-IDE-setting-is-not-used-when-using-.editorconfig](https://youtrack.jetbrains.com/issue/IJPL-148288/Continuation-indent-size-from-IDE-setting-is-not-used-when-using-.editorconfig)
> > @thorsten.schoening:
> > Thanks for reporting this issue. It seems we should use the continuations setting from the IDE's code style setting for your case.
> > As a workaround, you could add ij_continuation_indent_size = 8 in .editorconfig. to fix this issue.
> 
> I changed "continuation_indent_size" to "ij_continuation_indent_size", and it worked well,maybe i should ask another issue

Thanks for the feedback, we also had separate configuration options for `IDEA` before, which were merged later

@returnToInnocence any context for it? (Or shall we add both of them?)

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DataTypeUtil.java
```diff
@@ -124,8 +122,7 @@ public static UUID parseUUID(String key, Object rawValue) {
                                                          rawValue.getClass()));
     }
 
-    private static Object parseSingleValue(String key, Object rawValue,
```
imbajin, line 127: maybe we could enhance the logic like this:
image


(Or use strategy mode/interface in future)

MuLeiSY2021: Sure, we can use source.type() to fix this problem. Using the strategy pattern or interfaces might make the code more complex, and since this function is not complex enough, it's unnecessary to apply them here. The switch statement is good enough.

.github/workflows/codeql-analysis.yml
```diff
@@ -34,17 +34,17 @@ jobs:
...
-      uses: actions/checkout@v3
-
+    - name: Fetch Code
+      uses: actions/checkout@v4
     - name: Setup Java JDK
       uses: actions/setup-java@v3
       with:
+        java-version: ${{ matrix.JAVA_VERSION }}
         distribution: 'zulu'
-        java-version: '8'
```
imbajin, line 47: Note: client module still needs support for java8

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DataTypeUtil.java
```diff
@@ -124,63 +121,90 @@ public static UUID parseUUID(String key, Object rawValue) {
...
-            E.checkState(source instanceof FileSource,
-                         "Only accept FileSource when convert String value " +
-                         "to Date, but got '%s'", source.getClass().getName());
-            String dateFormat = ((FileSource) source).dateFormat();
-            String timeZone = ((FileSource) source).timeZone();
+        }
+        switch (dataType) {
+            case TEXT:
+                if (!isString) {
+                    value = rawValue.toString();
```
imbajin, line 138: seems the logic is redundant?

`TEXT -> value.toString()` is enough?

imbajin (approved):
Thanks for your contribution~

simon824 (approved):

MuLeiSY2021:
we also need close #641 ,or be permanent job?

PR #659, opened 2025-04-25
Metta writer

dawit-melka:
- implemented meTTa writer
- make updates to integrate the writer


imbajin:
Thanks for your contribution, but could u give more context about it?

PR #683, opened 2025-09-16
refactor(loader): support concurrent readers, short-id & Graphsrc

sadwitdastreetz:
## Purpose of the PR

This PR is a part of **updating HugegraphLoader to 2.0** and most importantly, it is NOT ready yet. It introduces a major refactor and enhancement to the HugeGraph Loader, aiming to improve **parallelism, stability, and compatibility** during data loading.  
It includes:
- More flexible file/HDFS source support
- Stream-based JDBC fetching (especially for Oracle/MySQL large tables)
- Short-ID support with schema proxy injection
- Graph-to-Graph schema synchronization
- Improved concurrency model for input progress tracking

These changes address issues with performance bottlenecks, Kerberos token expiration, Oracle missing rows, and lack of schema compatibility when importing from another graph.

## Main Changes

  1. Concurrency Testing: Verify thread safety under high load
  2. Graph Source Testing: Validate Graph-to-Graph migration scenarios
  3. Error Recovery: Test failure handling under various error conditions
  4. Performance Benchmarking: Compare throughput with previous implementation
  5. Resource Leak Testing: Ensure proper cleanup under error conditions
  
### Loader

  Refactored HugeGraphLoader with concurrent loading, Graph source support, and improved error handling.

  Major Changes

  Concurrency

  - Added multi-threaded loading with ExecutorService and CompletableFuture
  - Configurable parallelism via parallelCount option
  - Scatter mode for distributed loading

  Graph Source Support

  - Added Graph-to-Graph data migration
  - Schema replication from source graphs
  - Vertex/edge label migration with property filtering
  - Index replication with field validation

  Schema Management

  - Automatic graph creation with backend configuration
  - Short ID configuration support
  - Enhanced Groovy script execution

  Error Handling

  - Improved ServerException handling
  - Better exception chain preservation
  - Modified exception propagation behavior

  API Changes

  - load() always returns true instead of context.noError()
  - Added checkGraphExists() and setGraphMode() methods
  - Commented out reader.confirmOffset() in loadStruct()
  - Simplified main() error handling

  Breaking Changes

  - Exception handling behavior modified
  - Resource cleanup sequence changed
  - Return value semantics changed in load()

### Source Layer
- **FileSource**: added `dir_filter`, `extra_date_formats`, `headerCaseSensitive`, and `splitCount` for flexible directory/file filtering and single-file parallel reading.
- **HDFSSource**: 
  - Kerberos auto-renewal via scheduled task.
  - Replaced prefix matching with `FileFilter + DirFilter`, supports recursive directory traversal.
  - Unified error handling with `LoadException`.
- **GraphSource (new)**: sync schema directly from another HugeGraph instance.

### Reader Layer
- **FileReader**: refactored `init()` and moved scan logic into `split()` → multiple sub-readers per file.
- **HDFSFileReader**: fixed subdirectory traversal bug (using wrong variable), added `DirFilter`.
- **JDBCReader**: replaced `RowFetcher` with streaming `JDBCFetcher` to avoid Oracle data loss and improve performance.
- **FineLineFetcher**: added null checks to avoid NPE.

### Progress Layer
- **InputProgress**: refactored from single `loadingItem` to `Map<String, InputItemProgress>` for multi-file concurrent tracking.  
  - Thread-safety improvements with synchronized maps.  
  - New `markLoaded(Readable, boolean)` API for fine-grained progress confirmation.

### Filter Layer
- **ShortId support**:
  - Added `ShortIdParser`, `ShortIdConfig`.  
  - Schema enhancement via `SchemaManagerProxy` and `VertexLabelProxy` using reflection, injecting short-id handling transparently into HugeClient.

### Options
- Extended `LoadOptions` with new cluster, graph, and loading optimization flags (`--scatter-sources`, `--short-id`, `--restore`, etc.).  
- Added `dumpParams()` to log all runtime parameters.

### Others
- Added `GlobalExecutorManager` for thread pool management.  
- Updated `FileLoadTest` to adapt to `InputProgress` refactor.

```mermaid
sequenceDiagram
  autonumber
  actor CLI as 用户(CLI)
  participant Loader as HugeGraphLoader
  participant Options as LoadOptions
  participant Ctx as LoadContext
  participant Exec as GlobalExecutorManager
  participant Reader as InputReader(s)
  participant Parse as ElementParseGroup
  participant ShortId as ShortIdParser
  participant Client as HugeClient

  CLI->>Loader: new(args) / load()
  Loader->>Options: 解析并设置并行/shortId 等
  Loader->>Ctx: init(Options) -> 创建 indirectClient, filterGroup
  Loader->>Exec: getExecutor(parallel)
  Loader->>Reader: create reader / split() (若 multiReaders)
  par 并行处理每个 InputTaskItem
    Reader-->>Loader: emit Line 或 GraphElement
    Loader->>Parse: filter(element)
    Parse-->>Loader: true / false
    alt 通过过滤
      Loader->>ShortId: 可能转换短ID
      ShortId-->>Loader: 更新 element.id
      Loader->>Client: 写入（提交/flush 由配置或每行触发）
    else 过滤掉
      Loader-->>Loader: 跳过
    end
    Loader->>Ctx: 标记进度 (loaded_items/loading_items map)
  end
  Loader->>Exec: shutdown()
  Loader-->>CLI: 返回完成/异常

```

```mermaid
sequenceDiagram
  autonumber
  participant GSrc as GraphSource
  participant GReader as GraphReader
  participant Fetch as GraphFetcher
  participant Client as HugeClient
  participant Loader as HugeGraphLoader

  GSrc->>GReader: new(GraphSource)
  GReader->>Client: createHugeClient()
  loop 批量拉取
    GReader->>Fetch: queryBatch(offset, size)
    Fetch->>Client: 执行查询
    Client-->>Fetch: 返回元素批次
    Fetch-->>GReader: elements
    GReader-->>Loader: 生成 Line（交由 NopBuilder/过滤链处理）
  end

```

- [x] Modify configurations (`LoadOptions` extended)
- [x] The public API (Loader command line options & progress API)
- [ ] Dependencies
- [ ] Other impacts

## Documentation Status

- [x] `Doc - TODO` (need to update loader usage doc)

### DIFF
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
@@ -113,6 +113,7 @@ public final class HugeGraphLoader {
             loader = new HugeGraphLoader(args);
         } catch (Throwable e) {
             Printer.printError("Failed to start loading", e);
+            System.exit(1);
             return;
         }
 
@@ -178,12 +179,14 @@ public final class HugeGraphLoader {
                 () -> this.mapping.structs().stream().filter(struct -> !struct.skip())
                                   .map(InputStruct::input);
 
-        if (inputsSupplier.get().anyMatch(input -> SourceType.GRAPH.equals(input.type()))) {
-            if (!inputsSupplier.get().allMatch(input -> SourceType.GRAPH.equals(input.type()))) {
-                throw new LoadException("All inputs must be of Graph Type");
-            }
-            this.context().setRestoreMode();
-        } else if (this.options.restore) {
+        boolean allMatch = inputsSupplier.get().allMatch(input -> SourceType.GRAPH.equals(input.type()));
+        boolean anyMatch = inputsSupplier.get().anyMatch(input -> SourceType.GRAPH.equals(input.type()));
+
+        if (anyMatch && !allMatch) {
+            throw new LoadException("All inputs must be of Graph Type");
+        }
+
+        if (allMatch || this.options.restore) {
             this.context().setRestoreMode();
         } else {
             this.context().setLoadingMode();
@@ -219,11 +222,7 @@ public final class HugeGraphLoader {
                 LOG.warn(logMessage);
             }
 
-            RuntimeException e = LoadUtil.targetRuntimeException(t);
-            Printer.printError("Failed to load", e);
-            LOG.error("Load failed with exception", e);
-
-            throw e;
+            throw LoadUtil.targetRuntimeException(t);
         }
 
         return true;
@@ -354,7 +353,7 @@ public final class HugeGraphLoader {
             HugeClient sourceClient,
             HugeClient targetClient,
             List<? extends SchemaLabel> labels, // VertexLabel or EdgeLabel
-            Map<String, GraphSource.SeletedLabelDes> selectedMap,
+            Map<String, GraphSource.SelectedLabelDes> selectedMap,
             Map<String, GraphSource.IgnoredLabelDes> ignoredMap,
             boolean isVertex) {
 
@@ -419,10 +418,10 @@ public final class HugeGraphLoader {
             vertexLabels = sourceClient.schema().getVertexLabels();
         }
 
-        Map<String, GraphSource.SeletedLabelDes> mapSelectedVertices
+        Map<String, GraphSource.SelectedLabelDes> mapSelectedVertices
                 = new HashMap<>();
         if (graphSource.getSelectedVertices() != null) {
-            for (GraphSource.SeletedLabelDes des :
+            for (GraphSource.SelectedLabelDes des :
                     graphSource.getSelectedVertices()) {
                 mapSelectedVertices.put(des.getLabel(), des);
             }
@@ -474,10 +473,10 @@ public final class HugeGraphLoader {
             edgeLabels = sourceClient.schema().getEdgeLabels();
         }
 
-        Map<String, GraphSource.SeletedLabelDes> mapSelectedEdges
+        Map<String, GraphSource.SelectedLabelDes> mapSelectedEdges
                 = new HashMap<>();
         if (graphSource.getSelectedEdges() != null) {
-            for (GraphSource.SeletedLabelDes des :
+            for (GraphSource.SelectedLabelDes des :
                     graphSource.getSelectedEdges()) {
                 mapSelectedEdges.put(des.getLabel(), des);
             }
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/builder/EdgeBuilder.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/builder/EdgeBuilder.java
@@ -46,7 +46,7 @@ public class EdgeBuilder extends ElementBuilder<Edge> {
     private final VertexLabel sourceLabel;
     private final VertexLabel targetLabel;
     private final Collection<String> nonNullKeys;
-    // Used to optimize access performace
+    // Used to optimize access performance
     private VertexIdsIndex vertexIdsIndex;
     private String[] lastNames;
 
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/reader/file/FileLineFetcher.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/reader/file/FileLineFetcher.java
@@ -84,7 +84,7 @@ public class FileLineFetcher extends LineFetcher {
 
     @Override
     public void resetReader() {
-        LOG.error("resetReader called, reader reset to null, offset={}", this.offset());
+        LOG.debug("resetReader called, reader reset to null, offset={}", this.offset());
         this.reader = null;
     }
 
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/reader/graph/GraphReader.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/reader/graph/GraphReader.java
@@ -99,7 +99,7 @@ public class GraphReader extends AbstractReader {
         // Do with Vertex
         // 1. Get All Selected Vertex
         if (this.source.getSelectedVertices() != null) {
-            for (GraphSource.SeletedLabelDes selected :
+            for (GraphSource.SelectedLabelDes selected :
                     this.source.getSelectedVertices()) {
 
                 selectedVertices.put(selected.getLabel(), null);
@@ -141,7 +141,7 @@ public class GraphReader extends AbstractReader {
         // Do with edges
         // 1. Get All Selected Edges
         if (this.source.getSelectedEdges() != null) {
-            for (GraphSource.SeletedLabelDes selected :
+            for (GraphSource.SelectedLabelDes selected :
                     this.source.getSelectedEdges()) {
                 selectedEdges.put(selected.getLabel(), null);
                 if (selected.getQuery() != null && selected.getQuery().size() > 0) {
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/source/graph/GraphSource.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/source/graph/GraphSource.java
@@ -56,13 +56,13 @@ public class GraphSource extends AbstractSource {
     private String password;
 
     @JsonProperty("selected_vertices")
-    private List<SeletedLabelDes> selectedVertices;
+    private List<SelectedLabelDes> selectedVertices;
 
     @JsonProperty("ignored_vertices")
     private List<IgnoredLabelDes> ignoredVertices;
 
     @JsonProperty("selected_edges")
-    private List<SeletedLabelDes> selectedEdges;
+    private List<SelectedLabelDes> selectedEdges;
 
     @JsonProperty("ignored_edges")
     private List<IgnoredLabelDes> ignoredEdges;
@@ -97,7 +97,7 @@ public class GraphSource extends AbstractSource {
     }
 
     @Data
-    public static class SeletedLabelDes {
+    public static class SelectedLabelDes {
         @JsonProperty("query")
         private Map<String, Object> query;
 
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DataTypeUtil.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DataTypeUtil.java
@@ -63,7 +63,7 @@ public final class DataTypeUtil {
     public static Object convert(Object value, PropertyKey propertyKey,
                                  InputSource source) {
         E.checkArgumentNotNull(value, "The value of Property(%s) to be " +
-                                      "converted. can't be null", propertyKey.name());
+                                      "converted can't be null", propertyKey.name());
 
         String key = propertyKey.name();
         DataType dataType = propertyKey.dataType();
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/HugeClientHolder.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/HugeClientHolder.java
@@ -152,7 +152,7 @@ public final class HugeClientHolder {
         List<String> urls = clientFactory.getAutoURLs(options.cluster,
                                                       options.graphSpace, null);
 
-        E.checkState(CollectionUtils.isNotEmpty(urls), "No avaliable service!");
+        E.checkState(CollectionUtils.isNotEmpty(urls), "No available service!");
 
         int r = (int) Math.floor(Math.random() * urls.size());
         String url = urls.get(r);


hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -59,14 +88,34 @@ public final class HugeGraphLoader {
     private final LoadContext context;
```
imbajin, line 89: **Error Handling Issue**: The main() method's error handling has been changed to swallow exceptions silently by returning instead of rethrowing. This makes debugging and error tracking difficult in production.

**Recommendation**: Either restore the original exception throwing behavior or implement proper error logging and exit codes:
```java
public static void main(String[] args) {
    HugeGraphLoader loader;
    try {
        loader = new HugeGraphLoader(args);
    } catch (Throwable e) {
        Printer.printError("Failed to start loading", e);
        System.exit(1);  // Exit with error code
    }
    try {
        loader.load();
    } catch (Throwable e) {
        System.exit(1);
    }
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +203,62 @@ public boolean load() {
...
-                throw e;
-            }
-        } finally {
-            this.stopThenShutdown();
+            e.printStackTrace();
+
+            throw e;
         }
-        return this.context.noError();
+
```
imbajin, line 227: **Breaking API Change**: The `load()` method now always returns `true` instead of `context.noError()`. This is a breaking change that makes it impossible for callers to determine if the load operation succeeded.

**Issue**: Even when errors occur and exceptions are caught, the method still returns true, providing false success indicators to calling code.

**Recommendation**: Restore the original return semantics:
```java
public boolean load() {
    // ... existing code ...
    return this.context.noError();
}
```

Or at minimum, ensure proper error propagation so exceptions aren't silently swallowed.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +203,62 @@ public boolean load() {
...
+        // 任务执行成功
```
imbajin: **Resource Management Issue**: The `finally` block that calls `this.stopThenShutdown()` has been removed from the `load()` method. This creates a serious resource leak risk.

**Problem**: If an exception occurs during loading, resources may not be properly cleaned up since `stopThenShutdown()` won't be called.

**Recommendation**: Restore the finally block to ensure proper resource cleanup:
```java
public boolean load() {
    try {
        // ... loading logic ...
    } catch (Throwable t) {
        // ... error handling ...
        throw e;
    } finally {
        this.stopThenShutdown();  // Always cleanup resources
    }
    return true;
}
```

Alternatively, ensure that callers explicitly invoke `shutdown()` in their own finally blocks, which is less robust.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -59,14 +88,34 @@ public final class HugeGraphLoader {
     private final LoadContext context;
     private final LoadMapping mapping;
     private final TaskManager manager;
+    private final LoadOptions options;
+
+    // load 任务执行线程池
+    private ExecutorService loadService;
```
imbajin: **Thread Safety Concern**: The `loadService` ExecutorService is created as an instance field but is only initialized in `loadStructs()`. This could lead to NullPointerException if methods are called in unexpected order.

**Recommendation**: Initialize the executor in the constructor or use lazy initialization with proper synchronization:
```java
private volatile ExecutorService loadService;

private synchronized ExecutorService getLoadService(int parallelCount) {
    if (loadService == null) {
        loadService = ExecutorUtil.newFixedThreadPool(parallelCount, "loader");
    }
    return loadService;
}
```

Also ensure proper shutdown in all code paths.

imbajin: **Code Quality**: Chinese comments should be replaced with English for consistency with the rest of the codebase.

```java
// TODO: Replace with English comments
// Example: "Execution thread pool for load tasks"
```

This applies to all Chinese comments throughout the PR (e.g., line 94, 226, 265, etc.).

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +203,62 @@ public boolean load() {
...
         } catch (Throwable t) {
+            this.context.occurredError();
+
+            if (t instanceof ServerException) {
+                ServerException e = (ServerException) t;
+                String logMessage =
+                        "Log ServerException: \n" + e.exception() + "\n";
+                if (e.trace() != null) {
+                    logMessage += StringUtils.join((List<String>) e.trace(),
+                                                   "\n");
```
imbajin, line 220: **Unsafe Type Cast**: Unchecked cast `(List<String>) e.trace()` could cause ClassCastException at runtime.

**Recommendation**: Use proper type checking or safe casting:
```java
Object trace = e.trace();
if (trace instanceof List) {
    List<?> traceList = (List<?>) trace;
    logMessage += traceList.stream()
                          .map(Object::toString)
                          .collect(Collectors.joining("\n"));
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +203,62 @@ public boolean load() {
...
+                }
+                LOG.warn(logMessage);
+            }
+
             RuntimeException e = LoadUtil.targetRuntimeException(t);
             Printer.printError("Failed to load", e);
```
imbajin: **Exception Handling Issue**: Adding `e.printStackTrace()` in production code is not recommended. Stack traces should be logged via the logging framework, not printed to stderr.

**Recommendation**: Remove the printStackTrace() call and rely on proper logging:
```java
RuntimeException e = LoadUtil.targetRuntimeException(t);
Printer.printError("Failed to load", e);
LOG.error("Load failed with exception", e);  // Use logger instead
throw e;
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
+        LOG.info("waiting for loading finish {}", loadTasks.size());
+        // wait for finish
+        try {
+            CompletableFuture.allOf(loadTasks.toArray(new CompletableFuture[0]))
+                             .join();
+        } catch (CompletionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof ParseException) {
+                throw (ParseException) cause;
+            } else if (cause instanceof LoadException) {
```
imbajin, line 707: **Concurrency Issue**: The exception handling in `asyncLoadStruct` wraps and re-throws exceptions, but the concurrent execution model using `CompletableFuture.allOf().join()` may not preserve exception context properly for all failed tasks.

**Issue**: When multiple tasks fail concurrently, only the first exception is propagated while others are lost.

**Recommendation**: Consider collecting all exceptions:
```java
List<Throwable> exceptions = new ArrayList<>();
for (CompletableFuture<Void> future : loadTasks) {
    try {
        future.join();
    } catch (CompletionException e) {
        exceptions.add(e.getCause() != null ? e.getCause() : e);
    }
}
if (!exceptions.isEmpty()) {
    LoadException aggregated = new LoadException("Multiple load failures occurred");
    exceptions.forEach(aggregated::addSuppressed);
    throw aggregated;
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
+        }
+        if (parallelCount <= 0) {
+            parallelCount = structs.size();
+        }
+
+        boolean scatter = this.context.options().scatterSources;
+
+        LOG.info("{} threads for loading {} structs, from {} to {} in {} mode",
+                 parallelCount, structs.size(), this.context.options().startFile,
+                 this.context.options().endFile,
```
imbajin, line 677: **Typo**: "sequencial" should be "sequential" in the log message.

```java
scatter ? "scatter" : "sequential"
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
-                reader.init(this.context, struct);
-                // Load data from current input mapping
-                this.loadStruct(struct, reader);
+
+            // Create and init InputReader
+            try {
+                LOG.info("Start loading: '{}'", struct);
+
+                InputReader reader = InputReader.create(struct.input());
+                List<InputReader> readerList = reader.multiReaders() ?
```
imbajin, line 613: **Resource Leak Risk**: Readers created in `prepareTaskItems()` may not be properly closed if an exception occurs during initialization or if tasks are filtered out.

**Issue**: When `curFile < startFile` or `curFile >= endFile`, the reader is closed, but if an exception occurs before reaching the close() call, the reader remains open.

**Recommendation**: Use try-with-resources or ensure proper cleanup:
```java
try (InputReader reader = InputReader.create(struct.input())) {
    List<InputReader> readerList = reader.multiReaders() ?
                                   reader.split() :
                                   ImmutableList.of(reader);
    // ... rest of logic
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
             } catch (InitException e) {
                 throw new LoadException("Failed to init input reader", e);
             }
+            curIndex += 1;
+        }
+        // sort by seqNumber to allow scatter loading from different sources
+        if (scatter) {
+            tasks.sort(new Comparator<InputTaskItem>() {
+                @Override
+                public int compare(InputTaskItem o1, InputTaskItem o2) {
```
imbajin: **Code Style**: Using anonymous inner class for simple comparison. Consider using a lambda expression for better readability:

```java
tasks.sort((o1, o2) -> {
    if (o1.structIndex == o2.structIndex) {
        return o1.seqNumber - o2.seqNumber;
    }
    return o1.structIndex - o2.structIndex;
});
```

Or even better, use `Comparator.comparing()`:
```java
tasks.sort(Comparator.comparingInt((InputTaskItem o) -> o.structIndex)
                     .thenComparingInt(o -> o.seqNumber));
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
+
+        return tasks;
+    }
+
+    private void loadStructs(List<InputStruct> structs) {
+        int parallelCount = this.context.options().parallelCount;
+        if (structs.size() == 0) {
+            return;
+        }
+        if (parallelCount <= 0) {
```
imbajin, line 669: **Potential Integer Overflow**: Using `structs.size()` directly for thread pool size could create an excessive number of threads if there are many structs.

**Risk**: Creating thousands of threads can cause OutOfMemoryError or severe performance degradation.

**Recommendation**: Add a reasonable upper limit:
```java
if (parallelCount <= 0) {
    parallelCount = Math.min(structs.size(), Runtime.getRuntime().availableProcessors() * 2);
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +592,141 @@ private void loadInputs(List<InputStruct> structs) {
...
+        boolean scatter = this.context.options().scatterSources;
+
+        LOG.info("{} threads for loading {} structs, from {} to {} in {} mode",
+                 parallelCount, structs.size(), this.context.options().startFile,
+                 this.context.options().endFile,
+                 scatter ? "scatter" : "sequencial");
+
+        this.loadService = ExecutorUtil.newFixedThreadPool(parallelCount,
+                                                           "loader");
+
```
imbajin, line 679: **Missing Null Check**: The code doesn't verify if `taskItems` is empty before proceeding with executor creation and task submission.

**Issue**: If all structs are skipped or filtered out, an empty thread pool will still be created unnecessarily.

**Recommendation**: Add early return after task preparation:
```java
List<InputTaskItem> taskItems = prepareTaskItems(structs, scatter);
if (taskItems.isEmpty()) {
    LOG.info("No tasks to execute after filtering");
    return;
}
```

hugegraph-loader/pom.xml
```diff
@@ -542,6 +542,11 @@
             <version>${kafka.testcontainer.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.parboiled</groupId>
+            <artifactId>parboiled-core</artifactId>
+            <version>1.1.8</version> <!-- 稳定版本，与 Gremlin 3.4.x 等组件兼容 -->
```
imbajin: **Dependency Version**: Adding a hard-coded dependency version with a Chinese comment is not best practice.

**Issues**:
1. Chinese comment should be in English
2. Version should be defined in the properties section for easier maintenance
3. Comment in POM is redundant - version management should be centralized

**Recommendation**: 
```xml
<dependency>
    <groupId>org.parboiled</groupId>
    <artifactId>parboiled-core</artifactId>
    <version>${parboiled.version}</version>
</dependency>
```

And add to properties:
```xml
<parboiled.version>1.1.8</parboiled.version>
```

hugegraph-loader/assembly/travis/conf/hugegraph.truststore
Thespica, line 1: need to keep this file?

sadwitdastreetz: > need to keep this file?

I think not. I will put it in .gitignore next time

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: System.exit() will cause the forked VM termination exception.
I would remove this process and replace it by using
`loader.shutdown()` and `GlobalExecutorManager.shutdown(loader.options.shutdownTimeout)`
ONLY

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: now the shutdown process will be occur in the `finally` block in `main()`

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -97,10 +155,51 @@ public LoadContext context() {
         return this.context;
     }
 
+    private void checkGraphExists() {
+        HugeClient client = this.context.indirectClient();
+        String targetGraph = this.options.graph;
+        if (this.options.createGraph) {
```
Thespica: two if statements can be merged.

hugegraph-loader/src/test/java/org/apache/hugegraph/loader/test/unit/LoadProgressTest.java
```diff
@@ -1,4 +1,6 @@
 /*
+ * Copyright 2017 HugeGraph Authors
+ *
```
Thespica: license

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/JsonUtil.java
```diff
@@ -1,4 +1,6 @@
 /*
+ * Copyright 2017 HugeGraph Authors
+ *
```
Thespica: license

sadwitdastreetz: global searched and cleared

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +272,286 @@ private void createSchema() {
...
+                for (ShortIdConfig config : options.shorterIDConfigs) {
+                    PropertyKey propertyKey = client.schema().propertyKey(config.getIdFieldName())
+                                                    .ifNotExist()
+                                                    .dataType(config.getIdFieldType())
+                                                    .build();
+                    client.schema().addPropertyKey(propertyKey);
+                }
+                groovyExecutor.execute(script, client);
+                List<VertexLabel> vertexLabels = client.schema().getVertexLabels();
+                for (VertexLabel vertexLabel : vertexLabels) {
```
imbajin, line 284: **Resource Leak Risk**: The `HugeClient sourceClient` is created but may not be properly closed if an exception occurs during vertex/edge/index label creation. Consider using try-with-resources for all client creation or ensure proper cleanup in finally blocks.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +207,63 @@ public boolean load() {
...
-        LOG.info("The graph '{}' has been cleared successfully", options.graph);
-
-        options.timeout = requestTimeout;
-        client.close();
+        try {
+            LOG.info("Prepare to clear the data of graph '{}'", options.graph);
+            client.graphs().clearGraph(options.graph, "I'm sure to delete all data");
+            LOG.info("The graph '{}' has been cleared successfully",
+                     options.graph);
+            options.timeout = requestTimeout;
```
imbajin: **Error Handling Issue**: The catch block catches `Throwable` and rethrows it without restoring the timeout. Move `options.timeout = requestTimeout` to a finally block to ensure it's always restored, regardless of exception type.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +594,133 @@ private void loadInputs(List<InputStruct> structs) {
...
+        if (parallelCount <= 0) {
+            parallelCount = structs.size();
+        }
+
+        boolean scatter = this.context.options().scatterSources;
+
+        LOG.info("{} threads for loading {} structs, from {} to {} in {} mode",
+                 parallelCount, structs.size(), this.context.options().startFile,
+                 this.context.options().endFile,
+                 scatter ? "scatter" : "sequential");
```
imbajin, line 678: **Thread Safety Concern**: The `loadService` ExecutorService is an instance field created in `loadStructs()`. If `load()` is called multiple times, this could leak executors. Consider: 1) Creating it in the constructor, or 2) Adding null check + cleanup before reassigning, or 3) Making it a local variable.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -233,15 +733,18 @@ private void loadStruct(InputStruct struct, InputReader reader) {
...
+        long batchStartTime = System.currentTimeMillis();
+
+        for (boolean finished = false; !finished; ) {
             if (this.context.stopped()) {
                 break;
             }
             try {
                 // Read next line from data source
                 if (reader.hasNext()) {
                     Line next = reader.next();
```
imbajin, line 764: **Commented Critical Code**: The `reader.confirmOffset()` call is commented out. This appears critical for progress tracking. If intentional, add a clear comment explaining why. Without this, data may be reprocessed on restart or progress reporting could be incorrect.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +207,63 @@ public boolean load() {
...
-                throw e;
-            }
-        } finally {
-            this.stopThenShutdown();
+            LOG.error("Load failed with exception", e);
+
+            throw e;
         }
+
         return this.context.noError();
```
imbajin: **Breaking API Change**: The `load()` method now always returns `true` instead of `this.context.noError()`. This breaks the API contract for callers that check the return value. Consider a deprecation period or document this breaking change prominently in release notes and migration guides.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -97,10 +155,50 @@ public LoadContext context() {
         return this.context;
     }
 
+    private void checkGraphExists() {
+        HugeClient client = this.context.indirectClient();
+        String targetGraph = this.options.graph;
+        if (this.options.createGraph
+            && !client.graphs().listGraph().contains(targetGraph)) {
+            Map<String, String> conf = new HashMap<>();
```
imbajin, line 163: **Hardcoded Configuration**: The graph configuration hardcodes backend="hstore", serializer="binary", etc. This should be configurable through LoadOptions or a config file to support different deployment scenarios (RocksDB, MySQL, PostgreSQL backends, etc.).

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +272,286 @@ private void createSchema() {
...
+
+        sourceClient.assignGraph(graphSource.getGraphSpace(),
+                                 graphSource.getGraph());
+
+        // Create Vertex Schema
+        List<VertexLabel> vertexLabels = new ArrayList<>();
+        if (graphSource.getSelectedVertices() != null) {
+            List<String> selectedVertexLabels =
+                    graphSource.getSelectedVertices()
+                               .stream().map((des) -> des.getLabel())
```
imbajin, line 409: **NPE Risk**: `graphSource.getSelectedVertices()` could return null but is used directly in stream operations. While a null check exists later, the initial stream creation will throw NPE if null. Add null-safe check: `if (graphSource.getSelectedVertices() != null) { ... }` before the stream.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +594,133 @@ private void loadInputs(List<InputStruct> structs) {
...
+
+                LOG.info("total {} found in '{}'", readerList.size(), struct);
+                tasks.ensureCapacity(tasks.size() + readerList.size());
+                int seq = 0;
+                for (InputReader r : readerList) {
+                    if (curFile >= this.context.options().startFile &&
+                        (this.context.options().endFile == -1 ||
+                         curFile < this.context.options().endFile)) {
+                        // Load data from current input mapping
+                        tasks.add(new InputTaskItem(struct, r, seq, curIndex));
```
imbajin, line 626: **Resource Cleanup Issue**: When readers are filtered out by the file range check (startFile/endFile), they are closed with `r.close()`. However, if an exception occurs during the loop before reaching these readers, earlier readers won't be cleaned up. Consider tracking all created readers in a list and using try-finally to ensure cleanup.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -233,15 +733,18 @@ private void loadStruct(InputStruct struct, InputReader reader) {
         ParseTaskBuilder taskBuilder = new ParseTaskBuilder(this.context, struct);
         final int batchSize = this.context.options().batchSize;
         List<Line> lines = new ArrayList<>(batchSize);
```
imbajin, line 754: **Race Condition Risk**: `InputProgress` is modified from multiple async threads via `markLoaded()` calls. While the PR mentions thread-safety improvements with synchronized maps, ensure ALL access patterns including reads are properly synchronized. Consider using `ConcurrentHashMap` or explicit synchronization blocks.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +594,133 @@ private void loadInputs(List<InputStruct> structs) {
...
+            if (cause instanceof ParseException) {
+                throw (ParseException) cause;
+            } else if (cause instanceof LoadException) {
+                throw (LoadException) cause;
+            } else if (cause != null) {
+                if (cause instanceof RuntimeException) {
+                    throw (RuntimeException) cause;
+                } else {
+                    throw new RuntimeException(cause);
+                }
```
imbajin, line 714: **Exception Handling Issue**: The catch blocks handle different exception types but may not execute the finally block cleanup (`cleanupEmptyProgress()` and `loadService.shutdown()`) if certain exceptions are thrown. Wrap the entire try-catch in another try-finally to guarantee cleanup or ensure all exception paths lead to proper resource cleanup.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: the client will be closed in the finally blocks in `main()`, included in the function `context.close()`

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: modified `loadService` to local variable in the method loadStructs

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: > **Commented Critical Code**: The `reader.confirmOffset()` call is commented out. This appears critical for progress tracking. If intentional, add a clear comment explaining why. Without this, data may be reprocessed on restart or progress reporting could be incorrect.

The reader.confirmOffset() call is not commented out in the current implementation.
It is invoked after each batch to confirm progress and avoid data reprocessing on restart.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: > **Breaking API Change**: The `load()` method now always returns `true` instead of `this.context.noError()`. This breaks the API contract for callers that check the return value. Consider a deprecation period or document this breaking change prominently in release notes and migration guides.

Already modified in last commit

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: Thanks for pointing this out — agreed.
I’ve added backend, serializer, and schedulerType fields to LoadOptions so that users can configure different HugeGraph backends and deployment modes without modifying the source code.
Default values remain hstore, binary, and distributed for backward compatibility.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: The null-check is already present

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: There would only be `String` type. So I think it is not necessary

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: Thanks for the comment.
InputProgress internally synchronizes all modifying operations and uses thread-safe collections (ConcurrentHashMap and synchronizedMap).
Therefore, there is no race condition risk here.

.gitignore
```diff
@@ -85,6 +85,7 @@ output/
 tree.txt
 *.versionsBackup
 .flattened-pom.xml
+hugegraph-loader/assembly/travis/conf/hugegraph.truststore
```
imbajin:
```suggestion
*.truststore
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,25 +592,150 @@ private void loadInputs(List<InputStruct> structs) {
...
+                        (this.context.options().endFile == -1 ||
+                         curFile < this.context.options().endFile)) {
+                        // Load data from current input mapping
+                        tasks.add(new InputTaskItem(struct, r, seq, curIndex));
+                    } else {
+                        r.close();
+                    }
+                    seq += 1;
+                    curFile += 1;
+                }
```
imbajin, line 632: ‼️ **资源泄漏风险**: 在 `prepareTaskItems` 方法中,当发生 `InitException` 异常时,已经创建的 `InputReader` 可能没有被正确关闭。

**问题**: 如果在循环中途抛出异常,之前成功创建的 reader 可能不会被关闭,导致资源泄漏。

**建议**: 使用 try-finally 确保异常时也能清理资源:
```java
List<InputReader> readers = new ArrayList<>();
try {
    InputReader reader = InputReader.create(struct.input());
    readers.add(reader);
    List<InputReader> readerList = reader.multiReaders() ? 
                                   reader.split() : 
                                   ImmutableList.of(reader);
    // ... rest of logic
} catch (InitException e) {
    // Clean up any created readers
    for (InputReader r : readers) {
        try { r.close(); } catch (Exception ex) { /* log */ }
    }
    throw new LoadException("Failed to init input reader", e);
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,25 +592,150 @@ private void loadInputs(List<InputStruct> structs) {
...
+
+        return tasks;
+    }
+
+    private void loadStructs(List<InputStruct> structs) {
+        int parallelCount = this.context.options().parallelCount;
+        if (structs.size() == 0) {
+            return;
+        }
+        if (parallelCount <= 0) {
```
imbajin, line 669: ‼️ **线程池未关闭**: 在 `loadStructs` 中创建的 `loadService` ExecutorService 从未被显式关闭。

**问题**: 即使任务完成或发生异常,线程池也不会自动关闭,导致线程泄漏和 JVM 无法正常退出。

**建议**: 使用 try-finally 确保线程池被正确关闭:
```java
ExecutorService loadService = ExecutorUtil.newFixedThreadPool(parallelCount, "loader");
try {
    // ... all loading logic
} finally {
    loadService.shutdown();
    try {
        if (!loadService.awaitTermination(60, TimeUnit.SECONDS)) {
            loadService.shutdownNow();
        }
    } catch (InterruptedException e) {
        loadService.shutdownNow();
        Thread.currentThread().interrupt();
    }
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,25 +592,150 @@ private void loadInputs(List<InputStruct> structs) {
...
+            item.reader.init(this.context, item.struct);
+            // Load data from current input mapping
+            loadTasks.add(
+                    this.asyncLoadStruct(item.struct, item.reader,
+                                         loadService));
+        }
+
+        LOG.info("waiting for loading finish {}", loadTasks.size());
+        // wait for finish
+        try {
```
imbajin: ⚠️ **并发安全问题**: `InputProgress` 的并发访问可能存在竞态条件。多个线程同时调用 `markLoaded` 时,内部的 Map 操作需要保证线程安全。

**建议**: 检查 `InputProgress` 实现是否使用了 `ConcurrentHashMap` 或适当的同步机制。如果使用普通 HashMap,需要改为:
```java
private final Map<String, InputItemProgress> progressMap = new ConcurrentHashMap<>();
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,25 +592,150 @@ private void loadInputs(List<InputStruct> structs) {
...
+                        }
+                    }
+                }
             }
+            curIndex += 1;
+        }
+        // sort by seqNumber to allow scatter loading from different sources
+        if (scatter) {
+            tasks.sort(Comparator.comparingInt((InputTaskItem o) -> o.structIndex)
+                     .thenComparingInt(o -> o.seqNumber));
```
imbajin, line 658: ⚠️ **整数溢出风险**: `parallelCount` 的默认值计算可能不合理。当 `structs.size()` 非常大时,创建过多线程会导致系统资源耗尽。

**当前代码**:
```java
if (parallelCount <= 0) {
    parallelCount = structs.size();  // 可能创建数百个线程
}
```

**建议**: 设置合理的上限:
```java
if (parallelCount <= 0) {
    parallelCount = Math.min(structs.size(), 
                             Runtime.getRuntime().availableProcessors() * 2);
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,25 +592,150 @@ private void loadInputs(List<InputStruct> structs) {
...
+
+        ExecutorService loadService = ExecutorUtil.newFixedThreadPool(parallelCount,
+                                                                      "loader");
+
+        List<InputTaskItem> taskItems = prepareTaskItems(structs, scatter);
+        List<CompletableFuture<Void>> loadTasks = new ArrayList<>();
+
+        if (taskItems.isEmpty()) {
+            LOG.info("No tasks to execute after filtering");
+            return;
```
imbajin: ‼️ **错误传播不一致**: 在并发加载中,如果某个任务失败,其他任务会继续执行,可能导致部分数据加载成功、部分失败的不一致状态。

**问题**: `CompletableFuture.allOf().join()` 只会抛出第一个异常,但其他任务可能已经部分执行。

**建议**: 实现快速失败机制:
```java
AtomicBoolean hasFailed = new AtomicBoolean(false);
for (InputTaskItem item : taskItems) {
    if (hasFailed.get()) break;  // 快速失败
    loadTasks.add(
        this.asyncLoadStruct(item.struct, item.reader, loadService)
            .exceptionally(ex -> {
                hasFailed.set(true);
                // 取消其他任务
                loadTasks.forEach(f -> f.cancel(true));
                throw new CompletionException(ex);
            })
    );
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +270,286 @@ private void createSchema() {
...
+            Set<String> existedPKs =
+                    targetClient.schema().getPropertyKeys().stream()
+                                .map(pk -> pk.name()).collect(Collectors.toSet());
+
+            for (String pkName : vertexLabel.properties()) {
+                PropertyKey pk = sourceClient.schema()
+                                             .getPropertyKey(pkName);
+                if (!existedPKs.contains(pk.name())) {
+                    targetClient.schema().addPropertyKey(pk);
+                }
```
imbajin: ⚠️ **Schema 创建缺少事务性**: GraphSource 的 schema 创建过程(PropertyKey -> VertexLabel -> EdgeLabel -> IndexLabel)在失败时可能导致部分 schema 已创建的不一致状态。

**建议**: 
1. 添加回滚机制或幂等性检查
2. 在 schema 创建前检查是否已存在
3. 考虑使用事务或原子操作

```java
// 添加幂等性检查
Set<String> existingPKs = targetClient.schema().getPropertyKeys()
                                     .stream().map(PropertyKey::name)
                                     .collect(Collectors.toSet());
if (!existingPKs.contains(pk.name())) {
    targetClient.schema().addPropertyKey(pk);
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +270,286 @@ private void createSchema() {
...
+    /**
+     * create schema like graphdb when source is graphdb;
+     *
+     * @param graphSource
+     */
+    private void createGraphSourceSchema(GraphSource graphSource) {
+
+        try (HugeClient sourceClient = graphSource.createHugeClient();
+             HugeClient client = HugeClientHolder.create(this.options, false)) {
+
```
imbajin: 🧹 **代码重复**: `createGraphSourceVertexLabel` 和 `createGraphSourceEdgeLabel` 方法有大量重复代码。

**建议**: 提取公共逻辑到私有辅助方法:
```java
private <T> void processSelectedLabels(
    List<T> labels,
    Map<String, SeletedLabelDes> selectedMap,
    Function<T, String> nameGetter,
    Function<T, List<String>> propertiesGetter) {
    // 公共的选择和过滤逻辑
}
```

这样可以减少约 100 行重复代码,提高可维护性。

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +270,286 @@ private void createSchema() {
...
+            }
         }
+
         this.context.updateSchemaCache();
     }
 
+    /**
+     * create schema like graphdb when source is graphdb;
+     *
+     * @param graphSource
```
imbajin, line 336: ⚠️ **HugeClient 资源泄漏**: `createGraphSourceSchema` 方法中创建的 `sourceClient` 和 `client` 使用 try-with-resources,但如果在 schema 创建过程中抛出异常,可能导致部分操作未完成。

**建议**: 添加异常处理和清理日志:
```java
try (HugeClient sourceClient = graphSource.createHugeClient();
     HugeClient client = HugeClientHolder.create(this.options, false)) {
    createGraphSourceVertexLabel(sourceClient, client, graphSource);
    createGraphSourceEdgeLabel(sourceClient, client, graphSource);
    createGraphSourceIndexLabel(sourceClient, client, graphSource);
} catch (Exception e) {
    LOG.error("Failed to create graph source schema for {}: {}", 
              graphSource.getGraph(), e.getMessage(), e);
    throw new LoadException("Schema creation failed", e);
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: The thread pool is actually shut down in the finally block `loadService.shutdown(),` so there’s no resource leak here.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: already used `ConcurrentHashMap` to ensure cocurrent-safe

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: already modified `parallelCount = Math.min(structs.size(), Runtime.getRuntime().availableProcessors() * 2);`

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: this set is already in used. So no need to change

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +206,63 @@ public boolean load() {
...
-                throw e;
-            }
-        } finally {
-            this.stopThenShutdown();
+            LOG.error("Load failed with exception", e);
+
+            throw e;
         }
+
         return this.context.noError();
```
imbajin: ‼️ **Critical: 异常处理逻辑变更可能导致错误被掩盖**

在 `load()` 方法中,ServerException 被捕获后只是记录日志,然后重新抛出。但在 catch 块中调用了 `this.context.occurredError()` 之后,仍然抛出异常,这可能导致:

1. 错误状态被标记,但错误本身仍在传播
2. `load()` 方法最后返回 `this.context.noError()`,但在异常路径中这行代码永远不会被执行
3. 与原有的返回值语义不一致(原代码注释说 'load() always returns true instead of context.noError()')

**建议:**
- 明确异常处理策略:是捕获并返回false,还是让异常继续传播?
- 如果要改变返回值语义,需要更新相关调用方的逻辑

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +595,152 @@ private void loadInputs(List<InputStruct> structs) {
...
+    private void loadStructs(List<InputStruct> structs) {
+        int parallelCount = this.context.options().parallelCount;
+        if (structs.size() == 0) {
+            return;
+        }
+        if (parallelCount <= 0) {
+            parallelCount = Math.min(structs.size(), Runtime.getRuntime().availableProcessors() * 2);
+        }
+
+        boolean scatter = this.context.options().scatterSources;
```
imbajin, line 673: ‼️ **Critical: 资源泄漏风险 - ExecutorService 未正确关闭**

在 `loadStructs()` 方法中创建了 ExecutorService:
```java
ExecutorService loadService = ExecutorUtil.newFixedThreadPool(parallelCount, "loader");
```

但在异常情况下,这个线程池可能不会被正确关闭。虽然有 finally 块调用 `loadService.shutdown()`,但:

1. 如果 `prepareTaskItems()` 抛出异常,ExecutorService 还未创建,但 finally 块会尝试访问它
2. shutdown() 之后没有 awaitTermination(),可能导致任务未完成就退出
3. 在任务执行过程中发生异常时,已提交的任务可能处于不确定状态

**建议:**
```java
ExecutorService loadService = null;
try {
    loadService = ExecutorUtil.newFixedThreadPool(parallelCount, "loader");
    // ... 其他逻辑
} finally {
    if (loadService != null) {
        loadService.shutdown();
        if (!loadService.awaitTermination(60, TimeUnit.SECONDS)) {
            loadService.shutdownNow();
        }
    }
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +595,152 @@ private void loadInputs(List<InputStruct> structs) {
...
+                             .join();
+        } catch (CompletionException e) {
+            Throwable cause = e.getCause();
+            if (cause instanceof ParseException) {
+                throw (ParseException) cause;
+            } else if (cause instanceof LoadException) {
+                throw (LoadException) cause;
+            } else if (cause != null) {
+                if (cause instanceof RuntimeException) {
+                    throw (RuntimeException) cause;
```
imbajin, line 711: ‼️ **Critical: 线程安全问题 - InputProgress 的并发访问**

代码中多个线程并发访问 `InputProgress`,虽然使用了 synchronized map,但在某些操作中仍存在竞态条件:

在 `loadStruct()` 方法中:
```java
loadTasks.add(CompletableFuture.supplyAsync(() -> {
    // ...
    this.context.updateProgress(item.struct, item.reader);
    // ...
}, loadService));
```

`updateProgress()` 内部可能涉及多步操作,单纯使用 synchronized map 不足以保证原子性。

**建议:**
1. 使用 `ConcurrentHashMap` 配合原子操作
2. 或在更高层次(方法级别)加锁
3. 确保 progress 更新的原子性,特别是在读-修改-写场景中

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +595,152 @@ private void loadInputs(List<InputStruct> structs) {
...
+                    break;
+                }
             } catch (InitException e) {
                 throw new LoadException("Failed to init input reader", e);
+            } finally {
+                Set<InputReader> usedReaders = tasks.stream()
+                                                    .map(item -> item.reader)
+                                                    .collect(Collectors.toSet());
+                for (InputReader r : readers) {
+                    if (!usedReaders.contains(r)) {
```
imbajin, line 644: ⚠️ **Important: Reader 资源清理逻辑不完善**

在 `prepareTaskItems()` 的 finally 块中:
```java
} finally {
    Set<InputReader> usedReaders = tasks.stream()
                                        .map(item -> item.reader)
                                        .collect(Collectors.toSet());
    for (InputReader r : readers) {
        if (!usedReaders.contains(r)) {
            try {
                r.close();
            } catch (Exception ex) {
                LOG.warn("Failed to close reader: {}", ex.getMessage());
            }
        }
    }
}
```

问题:
1. 只关闭了 "未使用" 的 reader,但使用中的 reader 在哪里关闭?
2. 如果 `reader.init()` 失败,reader 可能处于半初始化状态,仍需清理
3. 异常被吞掉(只记录warn),可能隐藏重要的资源释放失败

**建议:**
- 明确 reader 的生命周期管理责任
- 使用 try-with-resources 或确保在任务完成后统一清理
- 考虑是否需要一个 Reader 注册表来跟踪所有创建的 reader

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -200,27 +595,152 @@ private void loadInputs(List<InputStruct> structs) {
...
+                } else {
+                    throw new RuntimeException(cause);
+                }
+            } else {
+                throw e;
             }
+        } catch (Throwable t) {
+            throw t;
+        } finally {
+            // Shutdown service
```
imbajin, line 721: ⚠️ **Important: CompletableFuture 异常处理不完整**

在处理并发任务结果时:
```java
for (CompletableFuture<Void> future : loadTasks) {
    try {
        future.join();
    } catch (CompletionException e) {
        this.context.occurredError();
        throw LoadUtil.targetRuntimeException(e);
    }
}
```

问题:
1. 当一个任务失败时,其他正在执行的任务如何处理?应该继续还是取消?
2. 没有超时机制,如果某个任务hang住,会导致整个加载过程卡死
3. 只捕获了 CompletionException,其他异常(如 CancellationException)会如何处理?

**建议:**
```java
try {
    future.get(timeout, TimeUnit.SECONDS); // 使用 get 替代 join,支持超时
} catch (TimeoutException e) {
    future.cancel(true);
    throw new LoadException("Task timeout", e);
} catch (ExecutionException | InterruptedException e) {
    // 统一异常处理
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +271,288 @@ private void createSchema() {
...
+                                                    .dataType(config.getIdFieldType())
+                                                    .build();
+                    client.schema().addPropertyKey(propertyKey);
+                }
+                groovyExecutor.execute(script, client);
+                List<VertexLabel> vertexLabels = client.schema().getVertexLabels();
+                for (VertexLabel vertexLabel : vertexLabels) {
+                    ShortIdConfig config;
+                    if ((config = options.getShortIdConfig(vertexLabel.name())) != null) {
+                        config.setLabelID(vertexLabel.id());
```
imbajin, line 287: ⚠️ **Important: Schema 创建缺少事务性保证**

在 `createSchema()` 方法中,先添加 PropertyKey,再执行 Groovy 脚本,最后创建 IndexLabel:

```java
PropertyKey propertyKey = client.schema().propertyKey(config.getIdFieldName())
                                .ifNotExist()
                                .dataType(config.getIdFieldType())
                                .build();
client.schema().addPropertyKey(propertyKey);

groovyExecutor.execute(script, client);

// ... later
client.schema().addIndexLabel(indexLabel);
```

问题:
1. 如果中间步骤失败(如 groovyExecutor.execute()),已创建的 PropertyKey 不会回滚
2. 多次运行可能导致部分 schema 重复创建
3. 没有验证 Groovy 脚本的执行结果

**建议:**
- 在创建 schema 前检查是否已存在
- 提供清理/回滚机制
- 记录每步操作结果,便于问题排查

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +271,288 @@ private void createSchema() {
...
+        // Create Vertex Schema
+        List<VertexLabel> vertexLabels = new ArrayList<>();
+        if (graphSource.getSelectedVertices() != null) {
+            List<String> selectedVertexLabels =
+                    graphSource.getSelectedVertices()
+                               .stream().map((des) -> des.getLabel())
+                               .collect(Collectors.toList());
+
+            if (!CollectionUtils.isEmpty(selectedVertexLabels)) {
+                vertexLabels =
```
imbajin, line 413: ⚠️ **Important: Graph-to-Graph schema 同步缺少完整性验证**

在 `createGraphSourceLabels()` 中:
```java
Set<String> existedPKs =
        targetClient.schema().getPropertyKeys().stream()
                    .map(pk -> pk.name()).collect(Collectors.toSet());

for (String pkName : label.properties()) {
    PropertyKey pk = sourceClient.schema().getPropertyKey(pkName);
    if (!existedPKs.contains(pk.name())) {
        targetClient.schema().addPropertyKey(pk);
    }
}
```

问题:
1. 只检查了 PropertyKey 的名称,没有验证数据类型是否一致
2. 如果目标图中已存在同名但类型不同的 PropertyKey,会导致数据不兼容
3. 没有处理 PropertyKey 的其他属性(如 Cardinality)

**建议:**
- 比对 PropertyKey 的完整定义,包括 dataType、cardinality 等
- 如果存在不兼容的 schema,给出明确的错误提示
- 考虑添加强制覆盖选项

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -153,11 +271,288 @@ private void createSchema() {
...
+        }
+
+        for (VertexLabel label : vertexLabels) {
+            if (mapSelectedVertices.getOrDefault(label.name(),
+                                                 null) != null) {
+                List<String> selectedProperties = mapSelectedVertices.get(
+                        label.name()).getProperties();
+
+                if (selectedProperties != null) {
+                    label.properties().clear();
```
imbajin, line 437: 🧹 **Minor: 代码重复 - createGraphSourceLabels 可以进一步简化**

在 `createGraphSourceVertexLabel` 和 `createGraphSourceEdgeLabel` 中存在大量重复代码:

```java
Map<String, GraphSource.SeletedLabelDes> mapSelectedVertices = new HashMap<>();
if (graphSource.getSelectedVertices() != null) {
    for (GraphSource.SeletedLabelDes des : graphSource.getSelectedVertices()) {
        mapSelectedVertices.put(des.getLabel(), des);
    }
}
```

**建议:**
提取公共方法:
```java
private <T extends GraphSource.LabelDes> Map<String, T> toLabelMap(List<T> labels) {
    if (labels == null) return Collections.emptyMap();
    return labels.stream().collect(Collectors.toMap(
        GraphSource.LabelDes::getLabel, 
        Function.identity()
    ));
}
```

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
```diff
@@ -109,42 +206,63 @@ public boolean load() {
...
 
-        options.timeout = requestTimeout;
-        client.close();
+        try {
+            LOG.info("Prepare to clear the data of graph '{}'", options.graph);
+            client.graphs().clearGraph(options.graph, "I'm sure to delete all data");
+            LOG.info("The graph '{}' has been cleared successfully",
+                     options.graph);
+        } catch (Exception e) {
+            LOG.error("Failed to clear data for graph '{}': {}", options.graph, e.getMessage(), e);
```
imbajin, line 250: 🧹 **Minor: 日志级别不一致**

在 `clearAllDataIfNeeded()` 中:
```java
} catch (Exception e) {
    LOG.error("Failed to clear data for graph '{}': {}", options.graph, e.getMessage(), e);
    throw e;
}
```

而在其他地方使用 LOG.warn()。建议统一关键操作失败的日志级别。

同时,既然已经在抛出异常前记录了 error 日志,上层再次捕获时可以考虑不重复记录,避免日志冗余。

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: no need to use `this.context.noError` no more, return true is fine

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: 使用中的reader在asyncLoadStruct中清理

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/HugeGraphLoader.java
sadwitdastreetz: It may not be necessary

hugegraph-loader/src/test/java/org/apache/hugegraph/loader/test/functional/LoadTest.java
```diff
@@ -45,7 +48,9 @@ public class LoadTest {
     protected static final String HTTPS_PROTOCOL = "https";
     protected static final String TRUST_STORE_PATH = "assembly/travis/conf/hugegraph.truststore";
     protected static final String FILE_URL = CommonUtil.PREFIX + "hugegraph.truststore";
-    protected static final HugeClient CLIENT = HugeClient.builder(URL, GRAPH).build();
+    protected static final HugeClient CLIENT = HugeClient.builder(URL, GRAPH)
+                                                         .configUser("admin", "pa")
```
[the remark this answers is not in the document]
sadwitdastreetz: it is just test situation, no need to be that complicated

Thespica (approved):
LGTM, Thank you @sadwitdastreetz !

imbajin (approved):

PR #632, opened 2024-10-20
BREAKING CHANGE(hubble): introduce new version V2

FrostyHec:
## 📌 Background & Motivation

Apache HugeGraph Hubble 2.0 is a major platform evolution aimed at modernizing the web UI, streamlining backend services, aligning core client drivers, and establishing robust continuous integration gates. 

This PR consolidates **370+ fragmented historic commits into a linear sequence of 47 clean, Conventional-compliant commits**. It purges legacy modules, introduces React + TS foundations, stabilizes E2E automation pipelines, and enforces license and quality metrics.

---

## 🛠️ Key Architectural Changes

The integration flow of Hubble 2.0 components can be represented below:

```mermaid
graph TD
    subgraph "Hubble 2.0 Web UI (hubble-fe)"
        FE_Pages["React + TS UI Components <br>(i18n / canvas / scheduler)"]
    end

    subgraph "Hubble 2.0 Backend (hubble-be)"
        BE_Controller["MVC Controllers <br>(auth / algorithm / task)"]
        BE_Service["GraphsService / SchemaService"]
        BE_H2["Metadata Database (H2)"]
    end

    subgraph "Client Layer (hugegraph-client)"
        HG_Client["HugeClient (BuilderFacade)"]
    end

    subgraph "Core Graph Engine"
        HG_Server["HugeGraph Server"]
    end

    %% Flow Relationships
    FE_Pages -- "Rest API Requests" --> BE_Controller
    BE_Controller -- "Persist State" --> BE_H2
    BE_Controller -- "Manage & Coordinate" --> BE_Service
    BE_Service -- "Interact" --> HG_Client
    HG_Client -- "Gremlin / Schema API" --> HG_Server
```

---

### 1. Frontend Retirement & Modernization (`hubble-fe`)
* **Legacy Cleanups**: Purged the obsolete `hubble-fe-old` directories (~50k lines of legacy code retired) to reduce package sizes and maintenance overhead.
* **i18n & Localization**: Standardized internationalization properties (`zh-CN` / `en-US`) across core workspaces (Olap components, import scheduler, vertex/edge modals, task panels).
* **UI Stabilization**: Refactored topological canvas controls, fixed double-triggering modal state resets during datasource creation, and localized canvas tooltips.

### 2. Backend Restructuring & API Alignment (`hubble-be`)
* **Facade Simplification**: Re-aligned `GraphsService` interface signatures to rely strictly on the standard `HugeClient` API instead of outdated custom wrappers.
[...]

PR #718, opened 2026-03-19
refactor(client): add graphRequired option in HugeClientBuilder

FrostyHec:
## Purpose of the PR

Due to the logic upgrade of the HugeGraph server, certain operations (such as authentication) no longer require a specific `graphspaces` or `graph`. However, some interfaces in `HugeClient` are currently incompatible and still necessitate these parameters during connection initialization. Therefore, an adaptation upgrade is required. Here, a backward-compatible upgrade solution was selected.

This pull request introduces a new option to bypass required argument checks when building a `HugeClientBuilder`, allowing for more flexible client initialization. The main changes include updating constructors and builder methods to support the `skipRequiredChecks` flag, modifying validation logic, and adding unit tests to verify this new behavior.

## Main Changes

Enhancements to client builder flexibility:

* Added a new `skipRequiredChecks` boolean parameter to the `HugeClientBuilder` constructor and the corresponding `HugeClient.builder` static method, allowing users to bypass mandatory checks for `url` and `graph` arguments when needed.
* Updated validation logic in both the constructor and the `build()` method of `HugeClientBuilder` to conditionally skip argument checks based on the `skipRequiredChecks` flag.

Testing:

* Introduced a new test class `HugeClientBuilderTest` with tests to ensure that the builder behaves correctly with and without the `skipRequiredChecks` flag, including cases where required arguments are missing.

## Verifying these changes

    - add new tests.

- [ ]  Nope
- [ ]  Modify configurations
- [x]  The public API

## Documentation Status

- [x]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`


imbajin (approved):
Merge it first, remember to apply in the hubble PR #632

PR #724, opened 2026-04-12
chore(loader): improve the perf in loader-ci

### LINKED ISSUE #723: [Improve][CI] Avoid repeated environment downloads in loader-ci
Observed while rerunning `loader-ci` during PR #716 review.

## Problem

The `Prepare env and service` step in `loader-ci` appears to spend a large amount of time repeatedly downloading or rebuilding external dependencies on each run, even when the versions do not change.

From the current workflow:

- `.github/workflows/loader-ci.yml` only caches `~/.m2`
- `hugegraph-loader/assembly/travis/install-hadoop.sh` always downloads `hadoop-2.8.5.tar.gz` from `archive.apache.org`
- `hugegraph-loader/assembly/travis/install-mysql.sh` always runs `docker pull mysql:5.7`
- `hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh` always clones `apache/hugegraph` and rebuilds the server package from source

The screenshot from the failing/re-run workflow shows `Prepare env and service` taking about 19 minutes, with a large Hadoop tarball download dominating the step.

```text
loader-ci
└─ Prepare env and service
   ├─ install-hadoop.sh
   │  └─ wget hadoop-2.8.5.tar.gz  (large tarball, repeated)
   ├─ install-mysql.sh
   │  └─ docker pull mysql:5.7     (repeated image pull)
   └─ install-hugegraph-from-source.sh
      └─ git clone + mvn package   (repeated source build)
```

## Why this matters

- CI duration is much longer than necessary
- CI becomes more fragile because it depends on multiple external downloads during the test phase
- Re-runs are expensive even when the code change is unrelated to loader integration environments
- Current cache coverage likely does not match the real bottlenecks

### Prefer official artifacts / containers over ad-hoc install scripts

- Replace the MySQL setup script with a GitHub Actions `services` container or another pinned official image
- Replace the Hadoop local install script with a pinned container/image or other official prebuilt artifact if possible
- For HugeGraph server, prefer a reusable prebuilt tarball/artifact for the pinned commit/version instead of cloning and packaging from source on every CI run

### If scripts must remain, make them cache-aware and idempotent

- Add cache coverage for downloaded tarballs or extracted runtime directories if we still use script-based setup
- Skip `wget` / `docker pull` / clone+build when the required artifact is already available
- Make the scripts check for existing files/directories before re-downloading or rebuilding
- Verify whether GitHub Actions cache is currently missing the relevant paths, or whether restore keys are ineffective for this use case

## Possible scope

- `.github/workflows/loader-ci.yml`
- `hugegraph-loader/assembly/travis/install-hadoop.sh`
- `hugegraph-loader/assembly/travis/install-mysql.sh`
- `hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh`

## Expected outcome

- Repeated `loader-ci` runs should not re-download the same Hadoop tarball every time
- MySQL setup should rely on a reusable/pinned container path rather than always pulling inside the script
- HugeGraph server setup should reuse a stable artifact or cacheable output where possible
- `Prepare env and service` time should drop significantly and become more stable

lokidundun:
## Purpose of the PR

- close #723  

## Main Changes

1. .github/workflows/loader-ci.yml
  - Migrated MySQL to GitHub Actions services: Adopted the official `mysql:8.0` container service with health check configured. This eliminates the need to execute `docker pull` and `docker run` for every workflow run.
  - Added Hadoop caching: Cached `~/hadoop-3.3.6.tar.gz` with a fixed cache key `hadoop-3.3.6`.
  - Added HugeGraph Server caching: Cached `~/hugegraph-cache-${{ env.COMMIT_ID }}`; the cache key contains the commit ID to enable auto-invalidation when the version changes.
  - Removed the invocation of `install-mysql.sh`: MySQL is now provisioned by GitHub Actions services.

2. hugegraph-loader/assembly/travis/install-hadoop.sh
  - Added local installation & cache validation: Skip re-download/extraction if `/usr/local/hadoop` already exists; use the cached tar.gz file directly if it exists (while the local directory is missing) to avoid repeated downloads from archive.apache.org.
  - Added writes to `GITHUB_ENV`: Ensure subsequent steps can correctly obtain `HADOOP_HOME` and `PATH` environment variables.
  - Made configuration writes idempotent: Only write `core-site.xml` and `hdfs-site.xml` if the configuration files do not exist or their content is mismatched.

3. hugegraph-loader/assembly/travis/install-mysql.sh
  - Added idempotent check: Directly skip `docker pull` and `docker run` if the `mysql:8.0` container is already running (e.g., provided by GitHub Actions services).

4. hugegraph-loader/assembly/travis/install-hugegraph-from-source.sh
  - Added build caching: Check if the packaged `apache-hugegraph-*.tar.gz` exists in `~/hugegraph-cache-${COMMIT_ID}` before building. Use the cached tarball directly and skip the time-consuming `git clone` + `mvn package` if it exists; copy the tarball to the cache directory for reuse in future runs after the build if it does not exist.
 
result:
image

## Verifying these changes

    - xxx

- [ ]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`


lokidundun:
@imbajin could you please take a look when you are convenient❤️

hugegraph-loader/assembly/travis/install-hadoop.sh
```diff
@@ -17,31 +17,55 @@
...
-echo "export HADOOP_COMMON_LIB_NATIVE_DIR=$HADOOP_HOME/lib/native" >> ~/.bashrc
-echo "export PATH=$PATH:$HADOOP_HOME/bin:$HADOOP_HOME/sbin" >> ~/.bashrc
+# Export for GitHub Actions subsequent steps
+if [[ -n "${GITHUB_ENV:-}" ]]; then
+    echo "HADOOP_HOME=${HADOOP_HOME}" >> "${GITHUB_ENV}"
+    echo "HADOOP_COMMON_LIB_NATIVE_DIR=${HADOOP_HOME}/lib/native" >> "${GITHUB_ENV}"
+    echo "PATH=${PATH}:${HADOOP_HOME}/bin:${HADOOP_HOME}/sbin" >> "${GITHUB_ENV}"
+fi
+
+echo "export HADOOP_HOME=${HADOOP_HOME}" >> ~/.bashrc
```
imbajin: ⚠️ The PR is making this script reusable across cached/repeated runs, but these unconditional `~/.bashrc` appends still make it non-idempotent: every invocation adds another Hadoop segment to `PATH`, so a reused runner or local repro environment keeps growing shell state. It would be safer to guard these writes (or rely on `GITHUB_ENV`/`GITHUB_PATH` in CI) so repeated runs stay stable.

```suggestion
if ! grep -qxF "export HADOOP_HOME=${HADOOP_HOME}" ~/.bashrc; then
    echo "export HADOOP_HOME=${HADOOP_HOME}" >> ~/.bashrc
fi
if ! grep -qxF "export HADOOP_COMMON_LIB_NATIVE_DIR=${HADOOP_HOME}/lib/native" ~/.bashrc; then
    echo "export HADOOP_COMMON_LIB_NATIVE_DIR=${HADOOP_HOME}/lib/native" >> ~/.bashrc
fi
if ! grep -qxF "export PATH=\$PATH:${HADOOP_HOME}/bin:${HADOOP_HOME}/sbin" ~/.bashrc; then
    echo "export PATH=\$PATH:${HADOOP_HOME}/bin:${HADOOP_HOME}/sbin" >> ~/.bashrc
fi
```

hugegraph-loader/assembly/travis/install-mysql.sh
```diff
@@ -23,12 +23,20 @@ TRAVIS_DIR=$(dirname $0)
 CONF=hugegraph-test/src/main/resources/hugegraph.properties
 MYSQL_USERNAME=root
 
-# Set MySQL configurations
+DB_NAME="${1:-mysql}"
+DB_PASS="${2:-root}"
+
+# Skip if MySQL is already running (e.g., provided by GitHub Actions service)
+if docker ps | grep -q "mysql:5.7"; then
```
imbajin: ⚠️ This broad container check changes the script semantics in a risky way. If a developer already has any unrelated `mysql:5.7` container running, the script now exits without ensuring the requested test container name/password/port match what the loader tests expect. 

That can turn a clean setup failure into a much harder-to-diagnose auth/connection failure later. Could we scope the idempotency check to the requested container instead?

```suggestion
if docker ps --format '{{.Names}}' | grep -qx "${DB_NAME}"; then
    echo "MySQL container ${DB_NAME} is already running, skipping setup."
    exit 0
fi
```

Also consider upgrading the mysql version to `8.x` (stable version) if possible

hugegraph-loader/assembly/travis/install-hadoop.sh
```diff
@@ -17,31 +17,55 @@
 #
 set -ev
 
-sudo wget https://archive.apache.org/dist/hadoop/common/hadoop-2.8.5/hadoop-2.8.5.tar.gz
+HADOOP_VERSION="2.8.5"
```
imbajin: Prefer to use 3.x (stable version)

lokidundun: done

hugegraph-loader/assembly/travis/install-mysql.sh
lokidundun: done

hugegraph-loader/assembly/travis/install-mysql.sh
```diff
@@ -23,12 +27,20 @@ TRAVIS_DIR=$(dirname $0)
 CONF=hugegraph-test/src/main/resources/hugegraph.properties
 MYSQL_USERNAME=root
```
imbajin, line 29: ‼️ This skip check only matches an exact container name. GitHub Actions service containers are usually not named `mysql`, so the guard will miss the case you want to avoid and the script will still try to start a second MySQL instance.

Consider checking whether port 3306 is already occupied or probing the database itself instead.

hugegraph-loader/src/main/java/org/apache/hugegraph/loader/source/jdbc/JDBCVendor.java
```diff
@@ -305,7 +305,8 @@ public String buildUrl(JDBCSource source) {
                   .setParameter("characterEncoding", Constants.CHARSET.name())
                   .setParameter("rewriteBatchedStatements", "true")
                   .setParameter("useServerPrepStmts", "false")
-                  .setParameter("autoReconnect", "true");
+                  .setParameter("autoReconnect", "true")
```
imbajin: ⚠️ `buildUrl()` is shared by MYSQL and POSTGRESQL. Putting a MySQL-only auth workaround in the common path leaks this flag into the PostgreSQL URL too.

Please scope `allowPublicKeyRetrieval` to the MySQL branch instead.

```suggestion
        uriBuilder.setParameter("useSSL", "false")
                  .setParameter("characterEncoding", Constants.CHARSET.name())
                  .setParameter("rewriteBatchedStatements", "true")
                  .setParameter("useServerPrepStmts", "false")
                  .setParameter("autoReconnect", "true");
```

imbajin (approved):
THX

PR #725, opened 2026-04-13
fix(hubble): reset upload quota by job and fix >10GB historical upload blockage

### LINKED ISSUE #496: [Bug] can't load files when > 10GB in hubble
### Problem Type (问题类型)

configs (配置项 / 文档相关)

### Before submit

- [X] 我已经确认现有的 [Issues](https://github.com/apache/hugegraph/issues) 与 [FAQ](https://hugegraph.apache.org/docs/guides/faq/) 中没有相同 / 重复问题 (I have confirmed and searched that there are no similar problems in the historical issue and documents)

### Environment (环境信息)

- Server Version: 1.0.0 (Apache Release Version)
- Backend: RocksDB x nodes, HDD or SSD 
- OS: xx CPUs, xx G RAM, Ubuntu 2x.x / CentOS 7.x 
- Data Size:  xx vertices, xx edges 

### Your Question (问题描述)

web界面导入文件，累计超过10G大小后，如何重置？

_No response_

_No response_

peakxy:
## Purpose of the PR

- close #496

This PR fixes the Hubble upload quota reset problem described in https://github.com/apache/hugegraph-toolchain/issues/496.
Previously, uploaded file size was effectively accumulated across historical jobs, and deleting an old import job did not fully release the occupied upload quota. As a result, once the historical uploaded files exceeded 10GB, users could no longer continue uploading files in the web UI or reset the quota by removing old jobs.

## Main Changes

- Change backend upload quota validation from global historical scope to current job scope by using `JobManager.jobSize` instead of relying on global file history.
- Add the `size` request parameter in the upload flow so the backend can validate the original file size instead of only seeing chunk sizes.
- Complete cascading cleanup when deleting an import job, including related `LoadTask` records, `FileMapping` metadata, and uploaded files on disk, so deleting an old job can actually release the quota.
- Align frontend cumulative size validation with backend behavior by counting uploaded files, queued local files, and newly selected files in the same job.
- Keep `total_size_bytes` as the raw byte field required by frontend quota calculation, remove the redundant `job_size_bytes` field, and deduplicate the related i18n message key.

## Verifying these changes

    - Create an import job in Hubble and upload files in multiple batches; verify that upload is allowed until the cumulative size of the current job exceeds 10GB.
    - Create a historical job with large uploaded files, then create a new job and upload a small file; verify that the new job is not blocked by historical jobs.
    - Delete an old job with uploaded files, then create a new job and continue uploading; verify that the quota is released after the old job is removed.
    - Run `git diff --check` to confirm there is no formatting issue in the current diff.

- [ ]  Nope
- [ ]  Modify configurations
- [ ]  The public API
- [x]  Other affects (Hubble data import upload quota validation, job cleanup flow, and related frontend/backend upload interaction)

## Documentation Status

- [ ]  `Doc - TODO`
- [x]  `Doc - Done`
- [ ]  `Doc - No Need`


hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/service/load/JobManagerService.java
```diff
@@ -132,4 +139,59 @@ public void remove(int id) {
...
+                        deleteDiskFiles(copiedMappings);
+                    }
+                });
+    }
+
+    private void deleteDiskFiles(List<FileMapping> mappings) {
+        for (FileMapping mapping : mappings) {
+            try {
+                this.fileMappingService.deleteDiskFile(mapping);
+            } catch (RuntimeException e) {
```
imbajin: ⚠️ By the time we get here, all `FileMapping` rows are already deleted and the transaction is committed. If `deleteDiskFile()` fails, we only log a warning and the orphaned upload directory becomes effectively untraceable, because there is no metadata left for a later retry/cleanup job to discover. Since this PR is meant to make job deletion actually free the uploaded data, please consider persisting a retryable cleanup record or surfacing the failure instead of silently losing the only reference to the disk path.

hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/controller/load/FileUploadController.java
```diff
@@ -262,28 +269,51 @@ private void checkFileValid(int connId, int jobId, JobManager jobEntity,
...
         // Check is there a file with the same name
         FileMapping oldMapping = this.service.get(connId, jobId, fileName);
         Ex.check(oldMapping == null ||
                  oldMapping.getFileStatus() == FileMappingStatus.UPLOADING,
                  "load.upload.file.existed", fileName);
+    }
+
+    private Long resolveSourceFileSize(MultipartFile file, Long fileSize,
+                                       int total, int index) {
+        if (fileSize != null) {
```
imbajin: ‼️ The quota check now trusts the client-provided `size` value as the source of truth. A crafted request can under-report that number and still upload a much larger file, because we never reconcile `sourceFileSize` with the bytes actually written/merged on the server before marking the upload successful. Please validate against a server-derived size (or re-check the merged file size) so the limit cannot be bypassed by tampering with the request parameter.

hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/controller/load/FileUploadController.java
```diff
@@ -262,28 +269,51 @@ private void checkFileValid(int connId, int jobId, JobManager jobEntity,
...
+            long singleFileSizeLimit = this.config.get(
+                    HubbleOptions.UPLOAD_SINGLE_FILE_SIZE_LIMIT);
+            Ex.check(sourceFileSize <= singleFileSizeLimit,
+                     "load.upload.file.exceed-single-size",
+                     FileUtils.byteCountToDisplaySize(singleFileSizeLimit));
+
+            long totalFileSizeLimit = this.config.get(
+                    HubbleOptions.UPLOAD_TOTAL_FILE_SIZE_LIMIT);
+            long currentJobSize = jobEntity.getJobSize();
+            Ex.check(sourceFileSize + currentJobSize <= totalFileSizeLimit,
```
imbajin: ‼️ `currentJobSize` only reflects completed files here, but Hubble uploads up to 5 files in parallel. If the job already has, for example, 9 GB uploaded, two concurrent 800 MB files can both pass this check against the same stale value and leave the job above the 10 GB cap once both merges finish. Please include in-progress uploads in the server-side reservation/check, or reserve the size before accepting the first chunk so the limit cannot be bypassed by concurrency.

peakxy:
@imbajin Hi, I have made revisions based on your comments. Please take a look again at your convenience. If there are still any issues, please point them out so I can make further changes. Sry, I'm a newcomer, so I appreciate your guidance.

imbajin:
补充一下 comment 里的几个 sec / CodeQL 提示，这边整体看了一遍。

这些告警大多是针对“用户输入参与路径拼接或文件操作”的通用提醒。就这次 PR 来说，核心目标还是修复 upload quota 的统计口径、分片上传时的大小校验，以及删除 job 后配额无法正确释放的问题。从目前代码来看，这条主线已经处理得比较完整了。

考虑到 Hubble 旧版本后续也不再持续维护，我觉得这些 sec 点不用作为这次 PR 的 blocker，更适合作为可选的防御性增强来看。建议只考虑下面几项：

1. 在 `FileUploadController` 里对 `name` 做一层基础校验，确保它就是纯文件名，拒绝包含 `/`、`\` 或 `..` 这样的值。
2. 在真正执行磁盘操作前，比如读取上传文件大小、删除路径时，再补一层 upload root 的约束，确认目标路径仍然位于上传目录下面。
3. 顺手修一下 `deleteUnfinishedFile()` 里 token 清理的小问题：现在用的是 `token.startsWith(mapping.getName())`，但 token 实际上是 `md5(fileName) + "-" + time`，这里更合理的是用 `md5(mapping.getName())` 去匹配前缀。

peakxy:
@imbajin 根据你的建议已经改好了~有空看一下

imbajin (approved):
THX, maybe could unify CI for the next step (merged the loader-ci improvement PR just now)

PR #732, opened 2026-05-02
chore(ci): replace hardcoded server commit with latest release

lokidundun:
## Purpose of the PR

- close #xxx  

## Main Changes

  1. CI: track latest HugeGraph stable release instead of hardcoded commit

  All 6 workflows (client-ci, client-go-ci, hubble-ci, loader-ci, spark-connector-ci, tools-ci) previously had a # TODO: replace it with the (latest - n) commit id (n >= 15)      
  reminder with a manually-pinned COMMIT_ID. These SHAs drift out of date and silently miss server-side regressions.

  Replaced the static value with an actions/github-script@v7 step that:
  - queries apache/hugegraph's latest GitHub release via getLatestRelease
  - resolves the tag through getRef and dereferences annotated tags via getTag (lightweight tags work too)
  - exports the 7-char SHA as COMMIT_ID for downstream install-hugegraph-from-source.sh $COMMIT_ID

  CI now follows the most recent stable release with zero manual upkeep.

  2. loader-ci: cache the built HugeGraph server

  Reused the resolved release SHA as a cache key (~/hugegraph-cache-${commit_id}) so subsequent CI runs skip the server build entirely as long as the upstream release hasn't      
  moved.

  3. hugegraph-client-go: adapt gremlin to HugeGraph 1.7.0+ graph-space bindings

  Server 1.7.0 introduced graph spaces and changed how the gremlin engine registers traversal sources — the graph is no longer a top-level Groovy global named <graph>, it's now   
  aliased as __g_DEFAULT-<graph> (where DEFAULT is the default graph space). The Java client was updated for this in GremlinManager.java; the Go client was not, so every gremlin  
  POST against a 1.7.0+ server failed with:

  Could not rebind [g] to [__g_hugegraph] as [__g_hugegraph] not in the Graph or TraversalSource global bindings

  Ported the Java behavior:

  - gemlin.go: aliases field on PostRequest / PostRequestData is now map[string]string (was an empty anonymous struct that silently dropped values).
  - Added buildDefaultAliases(transport) mirroring GremlinManager.java:46-56 and HugeClientBuilder.DEFAULT_GRAPHSPACE = "DEFAULT": returns {"graph": "DEFAULT-<g>", "g":
  "__g_DEFAULT-<g>"} when the caller hasn't supplied custom aliases.
  - PostRequest.Do auto-injects on every gremlin POST. GET is intentionally untouched — the Java client has no GET equivalent, and the server's /gremlin GET endpoint cannot accept
   JSON aliases query params due to a Jersey URI-template parser conflict on { / }.
  - gemlin_test.go: removed the GET assertion (would 500 on 1.7.0+ regardless of client changes), changed the POST script from hugegraph.traversal().V().limit(3) to g.V().limit(3)
   to match Java's g.V().count() convention, and surfaced server status/message on failure for easier diagnosis.

  ---

## Verifying these changes

    - xxx

- [ ]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [ ]  `Doc - No Need`

### DIFF
--- a/.github/actions/setup-java-env/action.yml
+++ b/.github/actions/setup-java-env/action.yml
@@ -25,13 +25,6 @@ runs:
         distribution: ${{ inputs.distribution }}
         cache: 'maven'
 
-    - name: Cache Maven packages
-      uses: actions/cache@v4
-      with:
-        path: ~/.m2/repository
-        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-        restore-keys: ${{ runner.os }}-maven-
-
     - name: Use staged maven repo settings
       if: ${{ inputs.use-stage == 'true' }}
       shell: bash
--- a/.github/actions/setup-java-env/action.yml
+++ b/.github/actions/setup-java-env/action.yml
@@ -1,0 +1,40 @@
+name: 'Setup Java and Maven Environment'
+description: 'Install JDK, cache Maven packages, and optionally apply staged Maven repository settings'
+
+inputs:
+  java-version:
+    description: 'Java version to install'
+    required: false
+    default: '11'
+  distribution:
+    description: 'JDK distribution'
+    required: false
+    default: 'zulu'
+  use-stage:
+    description: 'Whether to apply staged Maven repository settings'
+    required: false
+    default: 'true'
+
+runs:
+  using: composite
+  steps:
+    - name: Install JDK ${{ inputs.java-version }}
+      uses: actions/setup-java@v4
+      with:
+        java-version: ${{ inputs.java-version }}
+        distribution: ${{ inputs.distribution }}
+        cache: 'maven'
+
+    - name: Cache Maven packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: ${{ runner.os }}-maven-
+
+    - name: Use staged maven repo settings
+      if: ${{ inputs.use-stage == 'true' }}
+      shell: bash
+      run: |
+        cp $HOME/.m2/settings.xml /tmp/settings.xml
+        mv -vf .github/configs/settings.xml $HOME/.m2/settings.xml
--- a/.github/workflows/hubble-ci.yml
+++ b/.github/workflows/hubble-ci.yml
@@ -46,19 +46,21 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            const offset = 15;
-            const { data: commits } = await github.rest.repos.listCommits({
-              owner: 'apache',
-              repo: 'hugegraph',
-              per_page: offset + 1
+            const owner = 'apache';
+            const repo = 'hugegraph';
+            const { data: release } = await github.rest.repos.getLatestRelease({ owner, repo });
+            const { data: ref } = await github.rest.git.getRef({
+              owner, repo, ref: `tags/${release.tag_name}`
             });
-            if (commits.length <= offset) {
-              core.setFailed(`Not enough commits. Found ${commits.length}, needed ${offset + 1}`);
-              return;
+            let sha = ref.object.sha;
+            if (ref.object.type === 'tag') {
+              const { data: tag } = await github.rest.git.getTag({ owner, repo, tag_sha: sha });
+              sha = tag.object.sha;
             }
-            const sha = commits[offset].sha.substring(0, 7);
+            sha = sha.substring(0, 7);
             core.exportVariable('COMMIT_ID', sha);
-            console.log(`Using HugeGraph commit id: ${sha}`);
+            core.setOutput('commit_id', sha);
+            console.log(`Using HugeGraph release ${release.tag_name} (${sha})`);
 
       - name: Install JDK 11
         uses: actions/setup-java@v3
--- a/.github/workflows/hubble-ci.yml
+++ b/.github/workflows/hubble-ci.yml
@@ -23,9 +23,6 @@ on:
 
 env:
   TRAVIS_DIR: hugegraph-hubble/hubble-dist/assembly/travis
-  # TODO: replace it with the (latest - n) commit id (n >= 15)
-  # FIXME: hugegraph commit date: 2025-10-30
-  COMMIT_ID: 8c1ee71 # 5b3d295
 
 jobs:
   hubble-ci:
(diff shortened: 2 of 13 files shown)


.github/actions/setup-java-env/action.yml
```diff
@@ -0,0 +1,40 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Install JDK ${{ inputs.java-version }}
+      uses: actions/setup-java@v4
+      with:
+        java-version: ${{ inputs.java-version }}
+        distribution: ${{ inputs.distribution }}
+        cache: 'maven'
```
imbajin, line 26: ⚠️ **Redundant double Maven cache**

`actions/setup-java@v4` with `cache: 'maven'` already saves/restores `~/.m2/repository` internally using a `setup-java`-namespaced key. The explicit `Cache Maven packages` step below (lines 28-33) targets the **same directory** with a different key, so every job saves two separate cache entries for identical content.

For workflows that previously had only one cache mechanism (`tools-ci`, `spark-connector-ci`, `client-go-ci`, `loader-ci`), this PR silently doubles the cache storage and upload time.

Remove either:
- the `cache: 'maven'` line here, keeping the explicit step below, OR
- the entire `- name: Cache Maven packages` block (lines 28-33), relying solely on `setup-java`'s built-in caching.

Both approaches work; the latter is less YAML and the recommended pattern for `setup-java@v4`.

.github/workflows/loader-ci.yml
```diff
@@ -50,18 +47,16 @@ jobs:
         with:
           fetch-depth: 2
 
-      - name: Install JDK 11
-        uses: actions/setup-java@v4
+      - name: Get HugeGraph stable commit id
+        id: get-commit
+        uses: ./.github/actions/get-hugegraph-commit
```
imbajin, line 52: ⚠️ **Implicit `$COMMIT_ID` — invisible to readers of this file**

The `Prepare env and service` step (later in the workflow, not in the diff) still runs:
```sh
$TRAVIS_DIR/install-hugegraph-from-source.sh $COMMIT_ID
```
`COMMIT_ID` is no longer in the job `env` block. It works only because this composite action calls `core.exportVariable('COMMIT_ID', sha)` which writes to `$GITHUB_ENV`. A reader scanning the workflow file will see no `COMMIT_ID` definition and have no idea where it comes from.

All other workflows in this PR pass the SHA explicitly via `steps.get-commit.outputs.commit_id`. Align loader-ci to the same pattern — either convert the install step to use the `setup-hugegraph-server` composite action, or add an explicit `env` binding to that step:
```yaml
- name: Prepare env and service
  env:
    COMMIT_ID: ${{ steps.get-commit.outputs.commit_id }}
  run: |
    $TRAVIS_DIR/install-hadoop.sh
    $TRAVIS_DIR/install-hugegraph-from-source.sh $COMMIT_ID
```

.github/workflows/hubble-ci.yml
```diff
@@ -44,33 +41,23 @@ jobs:
         with:
           fetch-depth: 2
 
-      - name: Install JDK 11
-        uses: actions/setup-java@v3
+      - name: Get HugeGraph stable commit id
+        id: get-commit
+        uses: ./.github/actions/get-hugegraph-commit
```
imbajin, line 46: ⚠️ **Same implicit `$COMMIT_ID` issue as loader-ci**

The `Prepare env and service` step later in this file calls:
```sh
$TRAVIS_DIR/install-hugegraph.sh $COMMIT_ID
```
`steps.get-commit.outputs.commit_id` is never referenced by name anywhere in this file — it only works via the `core.exportVariable` side-effect. Make the dependency explicit:
```yaml
- name: Prepare env and service
  env:
    COMMIT_ID: ${{ steps.get-commit.outputs.commit_id }}
  run: |
    ...
    $TRAVIS_DIR/install-hugegraph.sh $COMMIT_ID
```

hugegraph-client-go/api/v1/gremlin/gemlin_test.go
```diff
@@ -31,22 +31,17 @@ func TestGremlin(t *testing.T) {
...
 
     respPost, err := client.Gremlin.Post(
-        client.Gremlin.Post.WithGremlin("hugegraph.traversal().V().limit(3)"),
+        client.Gremlin.Post.WithGremlin("g.V().limit(3)"),
     )
     if err != nil {
         log.Fatalln(err)
     }
+    if respPost.StatusCode != 200 {
+         t.Errorf("client.Gremlin.Post http_status=%d, gremlin_status=%d, message=%s",
```
imbajin: 🧹 **Extra leading space — fails `gofmt`**

The `t.Errorf` line has 5 spaces of indentation instead of a tab. `gofmt` will produce a diff on this file.

```suggestion
		t.Errorf("client.Gremlin.Post http_status=%d, gremlin_status=%d, message=%s",
			respPost.StatusCode, respPost.Data.Status.Code, respPost.Data.Status.Message)
```

imbajin (commented):
Overall the approach is solid: composite actions cleanly eliminate duplicated CI boilerplate, and the Go gremlin alias fix correctly addresses the 1.7.0 graph-space binding regression. Two issues need attention before merging, plus a couple of minor nits.

hugegraph-client-go/api/v1/gremlin/gemlin.go
```diff
@@ -132,6 +126,19 @@ func (g Post) WithGremlin(gremlin string) func(request *PostRequest) {
     }
 }
 
+func buildDefaultAliases(transport api.Transport) map[string]string {
+    cfg := transport.GetConfig()
+    graphSpace := cfg.GraphSpace
+    if graphSpace == "" {
+        graphSpace = "DEFAULT"
+    }
+    full := graphSpace + "-" + cfg.Graph
```
imbajin, line 138: 🧹 **Unguarded empty `Graph` field produces a silent bad alias**

If a caller constructs a client without setting `Graph`, `full` becomes `"DEFAULT-"` and the alias `g` becomes `"__g_DEFAULT-"`. The server will accept the POST but return an error like `"TraversalSource not found"`, which is very hard to diagnose.

Since `NewCommonClient` does not validate that `Graph` is non-empty, a small guard here makes failures explicit:
```suggestion
	if cfg.Graph == "" {
		cfg.Graph = "hugegraph"
	}
	full := graphSpace + "-" + cfg.Graph
```
(Or add the check to `NewCommonClient` and document the default.)

imbajin (approved):
THX~

PR #734, opened 2026-05-05
fix(loader): fix thread-safety issue in DateUtil with concurrent date parsing

### LINKED ISSUE #733: [Bug] DateUtil uses shared SimpleDateFormat instances causing thread-safety issues in concurrent date parsing
### Bug Type (问题类型)

data inconsistency (数据不一致)

`DateUtil.parse()` and `DateUtil.now()` methods are not thread-safe. When called concurrently from multiple threads with different time zones, the time zone settings can interfere with each other, leading to incorrect parsing results.

```java
Thread A: setTimeZone("GMT+8") → Thread B: setTimeZone("GMT+0") → Thread A: parse() // uses wrong GMT+0 timezone
```

SYaoJun:
## Purpose of the PR

- close #733   

## Main Changes

Replace ConcurrentHashMap with ThreadLocal<HashMap> so each thread has its
own isolated SafeDateFormat instances, eliminating race conditions.

Ensure DateUtil.now() explicitly sets the default timezone before
formatting to avoid timezone pollution from previous parse() calls.

## Verifying these changes

    - Dtest=DateUtilTest

- [x]  Nope
- [ ]  Modify configurations
- [ ]  The public API

## Documentation Status

- [ ]  `Doc - TODO` 
- [ ]  `Doc - Done` 
- [x]  `Doc - No Need`

### DIFF
--- a/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DateUtil.java
+++ b/hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DateUtil.java
@@ -18,16 +18,15 @@
 package org.apache.hugegraph.loader.util;
 
 import java.util.Date;
-import java.util.Map;
 import java.util.TimeZone;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.hugegraph.date.SafeDateFormat;
 import org.apache.hugegraph.loader.constant.Constants;
 
 public final class DateUtil {
 
-    private static final Map<String, SafeDateFormat> DATE_FORMATS = new ConcurrentHashMap<>();
+    private static final ThreadLocal<java.util.HashMap<String, SafeDateFormat>> DATE_FORMATS =
+            ThreadLocal.withInitial(java.util.HashMap::new);
 
     public static Date parse(String source, String df) {
         return parse(source, df, Constants.TIME_ZONE);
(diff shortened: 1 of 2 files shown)


hugegraph-loader/src/main/java/org/apache/hugegraph/loader/util/DateUtil.java
```diff
@@ -18,16 +18,15 @@
...
 import java.util.TimeZone;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.hugegraph.date.SafeDateFormat;
 import org.apache.hugegraph.loader.constant.Constants;
 
 public final class DateUtil {
 
-    private static final Map<String, SafeDateFormat> DATE_FORMATS = new ConcurrentHashMap<>();
+    private static final ThreadLocal<java.util.HashMap<String, SafeDateFormat>> DATE_FORMATS =
```
imbajin: ⚠️ Please avoid using fully-qualified JDK type names inline here. There is no naming conflict in this file, so `java.util.HashMap` should be imported and referenced as `HashMap` instead. The same applies to the method reference below. Keeping the fully-qualified name in the declaration makes the code harder to read and goes against the project's Java style conventions.

imbajin (approved):

PR #746, opened 2026-07-21
feat: add Codex Hubble development environment

imbajin:
## Summary

Add a repository-managed Codex environment for local Hubble development.

| Before | After |
| --- | --- |
| Developers manually coordinate Hubble processes and local HStore setup across worktrees. | Codex actions provide deterministic frontend ports, shared backend ownership, optional HStore profiles, and fast/full verification. |
| Hubble local-development documentation has no Codex entry point. | The Hubble README links to the optional Codex worktree workflow. |

## Scope

- Adds only repository-scoped environment scripts and Compose overlays.
- Keeps runtime ownership files under `~/.codex/run/`; no secrets or runtime artifacts are tracked.
- Low-memory HStore starts with local `latest` images and does not pull automatically.

## Verification

- `bash .codex/tests/environment_test.sh`
- `docker compose ... config --quiet` for low-memory and balanced profiles
- `git diff --check`

### DIFF
--- a/.codex/infra/compose.low-memory.yml
+++ b/.codex/infra/compose.low-memory.yml
@@ -1,3 +1,20 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 services:
   pd:
     pull_policy: never
@@ -72,6 +89,7 @@ services:
       - "--"
       - /opt/toolchain-low-memory/patch-server-config.sh
     environment:
+      PASSWORD: ${HUBBLE_SERVER_PASSWORD:-pa}
       JAVA_OPTS: >-
         -Xms128m -Xmx192m
         -XX:ActiveProcessorCount=4
--- a/.codex/infra/compose.low-memory.yml
+++ b/.codex/infra/compose.low-memory.yml
@@ -1,0 +1,90 @@
+services:
+  pd:
+    pull_policy: never
+    mem_limit: 512m
+    cpus: 2.0
+    environment:
+      JAVA_OPTS: >-
+        -Xms64m -Xmx64m
+        -XX:ActiveProcessorCount=2
+        -XX:MaxMetaspaceSize=96m
+        -XX:MaxDirectMemorySize=64m
+        -XX:+UseContainerSupport
+        -Dlog4j2.asyncLoggerConfigRingBufferSize=4096
+        -DAsyncLoggerConfig.RingBufferSize=4096
+        -Dthread.pool.grpc.core=8
+        -Dthread.pool.grpc.max=32
+        -Dthread.pool.grpc.queue=256
+        -Djob.interruptableThreadPool.core=1
+        -Djob.interruptableThreadPool.max=8
+        -Djob.interruptableThreadPool.queue=128
+        -Djob.uninterruptibleThreadPool.core=0
+        -Djob.uninterruptibleThreadPool.max=4
+        -Djob.uninterruptibleThreadPool.queue=64
+        -Dpartition.default-shard-count=1
+        -Dpartition.store-max-shard-count=12
+
+  store:
+    pull_policy: never
+    mem_limit: 512m
+    cpus: 2.0
+    environment:
+      JAVA_OPTS: >-
+        -Xms64m -Xmx64m
+        -XX:ActiveProcessorCount=2
+        -XX:MaxMetaspaceSize=96m
+        -XX:MaxDirectMemorySize=64m
+        -XX:+UseContainerSupport
+        -Dlog4j2.asyncLoggerConfigRingBufferSize=4096
+        -DAsyncLoggerConfig.RingBufferSize=4096
+        -Drocksdb.total_memory_size=67108864
+        -Drocksdb.write_buffer_size=1048576
+        -Drocksdb.min_write_buffer_number_to_merge=2
+        -Dthread.pool.grpc.core=8
+        -Dthread.pool.grpc.max=32
+        -Dthread.pool.grpc.queue=256
+        -Dthread.pool.scan.core=4
+        -Dthread.pool.scan.max=16
+        -Dthread.pool.scan.queue=128
+        -Djob.interruptableThreadPool.core=2
+        -Djob.interruptableThreadPool.max=8
+        -Djob.interruptableThreadPool.queue=128
+        -Djob.uninterruptibleThreadPool.core=0
+        -Djob.uninterruptibleThreadPool.max=4
+        -Djob.uninterruptibleThreadPool.queue=64
+        -Dquery.push-down.threads=8
+        -Dquery.push-down.fetch_batch=1000
+        -Dquery.push-down.fetch_timeout=30000
+        -Dquery.push-down.memory_limit_count=5000
+        -Dquery.push-down.index_size_limit_count=5000
+        -Draft.disruptorBufferSize=256
+        -Draft.metrics=false
+        -Draft.maxReplicatorInflightMsgs=16
+        -Draft.maxEntriesSize=32
+        -Draft.maxBodySize=131072
+
+  server:
+    pull_policy: never
+    mem_limit: 768m
+    cpus: 4.0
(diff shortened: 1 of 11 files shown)


.codex/scripts/hubble.sh
```diff
@@ -0,0 +1,473 @@
...
+    SERVICE_OWNER=${owner}
+    SERVICE_TOKEN=${token}
+    trap cleanup_claim EXIT
+
+    java_home=$(java11_home)
+    runner=$(maven_runner "${java_home}")
+    classpath="${RUNTIME_DIR}/hubble-be.classpath"
+    echo "[hubble] compiling BE with Java 11 + ${runner}"
+    (
+        cd "${WORKTREE}/hugegraph-hubble"
```
imbajin, line 388: ‼️ A clean Codex worktree cannot start this BE action. This build only includes `hubble-be`, while that module depends on the current `hugegraph-loader`/client artifacts and `loader_home()` later requires an ignored `hugegraph-loader/apache-hugegraph-loader-*` distribution. `setup.sh` only installs frontend dependencies, so a fresh clone has neither prerequisite (the existing Hubble CI explicitly installs client+loader first). Please build/install the required reactor modules and produce/select the current Loader distribution deterministically, or provision a declared artifact, and cover `be-start` from a clean checkout.

.codex/scripts/hubble.sh
```diff
@@ -0,0 +1,473 @@
...
+            if [[ "${owner_worktree}" == "${WORKTREE}" ]]; then
+                echo "[hubble] BE already runs from this worktree (pid ${pid})" >&2
+                return 10
+            fi
+            if [[ "${switch}" != true ]]; then
+                echo "ERROR: BE is owned by ${owner_worktree} (pid ${pid})" >&2
+                echo "Use the 'Hubble BE Switch' action to replace it." >&2
+                return 1
+            fi
+            rm -f "${owner}"
```
imbajin, line 340: ‼️ The switch path drops ownership before the service has actually stopped. `kill_tree()` only sends TERM; this loop waits for the wrapper shell PID, not the Java listener/descendants, so a slow graceful shutdown leaves an unowned listener and the subsequent port check rejects the switch. `stop_current()` has the same ownership-loss pattern without any wait. Please retain the owner until the captured process tree and port have exited (with a bounded timeout and non-zero failure), and add a delayed-TERM lifecycle test.

.codex/scripts/infra.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+infra_pull() {
+    local server_repo
+    server_repo=$(resolve_server_repo)
+    assert_docker
+    compose_files low "${server_repo}"
+    echo "[infra] explicitly updating latest images"
+    TOOLCHAIN_CODEX_DIR="${CODEX_DIR}" HUGEGRAPH_VERSION=latest \
+        docker compose "${COMPOSE_ARGS[@]}" pull
```
imbajin, line 176: ⚠️ `Infrastructure Pull` currently skips every image because the low-memory overlay sets `pull_policy: never`. On this head, `docker compose ... --dry-run pull` reports `Skipped` for server/store/PD, while adding `--policy always` reports the expected pulls. Please explicitly override the service policy here (for example, `pull --policy always`) and test that this action really refreshes the images.

.codex/infra/compose.low-memory.yml
```diff
@@ -0,0 +1,108 @@
...
+  server:
+    pull_policy: never
+    mem_limit: 768m
+    cpus: 4.0
+    entrypoint:
+      - /usr/bin/dumb-init
+      - "--"
+      - /opt/toolchain-low-memory/patch-server-config.sh
+    environment:
+      PASSWORD: ${HUBBLE_SERVER_PASSWORD:-pa}
```
imbajin, line 92: ⚠️ This known default administrator password is combined with ports published on every host interface. The merged Compose config has no `host_ip` for 8080, 8520, or 8620, so the documented `admin/pa` Server and the Store/PD management endpoints are reachable from the developer's LAN. Please bind the necessary Server port to loopback, avoid publishing Store/PD to the host by default, and require or safely generate a local password.

.codex/scripts/infra.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+infra_stop() {
+    local server_repo
+    server_repo=$(resolve_server_repo)
+    assert_docker
+    compose_files low "${server_repo}"
+    echo "[infra] stopping Compose services; named volumes are preserved"
+    TOOLCHAIN_CODEX_DIR="${CODEX_DIR}" HUGEGRAPH_VERSION=latest \
+        docker compose "${COMPOSE_ARGS[@]}" down
```
imbajin, line 145: ⚠️ This `down` does not preserve the Server state as documented. `hugegraph/server:latest` declares `VOLUME /hugegraph-server`, but the merged Compose model does not attach a named volume there; removing the container leaves an anonymous volume that the next `up` will not reuse, including the initialization marker and local configuration. Please declare a stable named Server volume (and let reset remove it), or use `docker compose stop` for the preserving operation, with a stop/start persistence test.

.codex/infra/compose.low-memory.yml
```diff
@@ -0,0 +1,108 @@
...
+        -XX:MaxDirectMemorySize=64m
+        -XX:+UseContainerSupport
+        -Dlog4j2.asyncLoggerConfigRingBufferSize=4096
+        -DAsyncLoggerConfig.RingBufferSize=4096
+        --add-exports=java.base/jdk.internal.reflect=ALL-UNNAMED
+        --add-modules=jdk.unsupported
+        --add-exports=java.base/sun.nio.ch=ALL-UNNAMED
+    volumes:
+      - type: bind
+        source: ${TOOLCHAIN_CODEX_DIR}/infra/patch-server-config.sh
```
imbajin, line 106: ⚠️ The shared HStore service persists beyond worktree cleanup, but this bind mount points into the worktree that launched it. Once that temporary checkout is deleted, a container restart/recreation can no longer mount or execute the patch script. Please copy the script to a stable runtime location such as `~/.codex/run/...`, or package it in a stable image/volume, before starting the persistent service.

.codex/tests/environment_test.sh
```diff
@@ -0,0 +1,185 @@
...
+
+    if output=$(HG_SERVER_CONF_DIR="${missing_yaml}" \
+        "${CODEX_DIR}/infra/patch-server-config.sh" --patch-only 2>&1); then
+        fail "patch unexpectedly accepted a missing gremlin-server.yaml"
+    fi
+    assert_contains "${output}" "ERROR: missing ${missing_yaml}/gremlin-server.yaml"
+    grep -qx 'batch.max_write_threads=16' "${missing_yaml}/rest-server.properties"
+}
+
+test_environment_toml
```
imbajin, line 176: ‼️ This new contract suite is not wired into any CI job. Repository workflows do not invoke `environment_test.sh`, and `hubble-ci.yml` does not watch `.codex/**`; the current green Hubble job ran because the README changed and never executed these tests. Please add a workflow/step triggered by `.codex/**` that runs this suite, shell/static checks, and both layered `docker compose config` validations, including the BE and infrastructure lifecycle cases above.

.codex/tests/environment_test.sh
```diff
@@ -0,0 +1,185 @@
...
+    local expected=$2
+    [[ "${actual}" == *"${expected}"* ]] ||
+        fail "expected '${actual}' to contain '${expected}'"
+}
+
+test_environment_toml() {
+    python3 - "${CODEX_DIR}/environments/environment.toml" <<'PY'
+import pathlib
+import sys
+import tomllib
```
imbajin, line 45: ⚠️ The advertised `bash .codex/tests/environment_test.sh` command is not runnable with the target Mac's default Python: `/usr/bin/python3` 3.9.6 fails here with `ModuleNotFoundError: No module named 'tomllib'`, while setup neither provisions nor checks Python 3.11+. Please use a compatible parser/fallback, or explicitly provision and validate Python 3.11+ in setup, documentation, and CI.

imbajin (commented):
Blocking: yes. Summary: The new environment has clean-worktree startup, lifecycle ownership, image refresh, storage persistence, and local exposure issues that should be fixed before merge. Evidence: six independent review lanes; bash syntax, ShellCheck, environment tests, Compose expansion/dry-run, and Docker image metadata.

.codex/scripts/hubble.sh
```diff
@@ -0,0 +1,473 @@
...
+    command=$(ps -o command= -p "${pid}" 2>/dev/null || true)
+    [[ "${command}" == *".codex/scripts/hubble.sh"* ]]
+}
+
+owner_service_matches() {
+    local file=$1 pid port listener
+    owner_identity_matches "${file}" || return 1
+    pid=$(read_field "${file}" pid)
+    port=$(read_field "${file}" port)
+    listener=$(lsof -tiTCP:"${port}" -sTCP:LISTEN 2>/dev/null | head -1 || true)
```
imbajin, line 87: ⚠️ Missing `lsof` is silently treated as 'no listener'; similarly, `kill_tree()` swallows a missing `pgrep`. On a host without these undeclared tools, the script can claim an occupied port or stop only the wrapper and orphan Java/Node children. Please validate these required commands during setup/start with an actionable error, or provide reliable portable fallbacks instead of converting command-not-found into an empty result.

PR #750, opened 2026-08-01
fix(hubble): cache CORS preflight responses

cragkhit:
Hi. We are researchers from Mahidol University, Thailand, and the State University of Ceará, Brazil, working on a research project for improving open-source projects by using the latest accepted answer from Stack Overflow that matched your code snippet. We found this recommendation for improving your code from https://stackoverflow.com/a/58370217.

Note: Our study is approved by the Institutional Review Board of Mahidol University. You can find the participant information sheet explaining this study https://drive.google.com/file/d/1ml5AqrtWQ9pnifTQyTFTcWQmwp6RuPA7/view?usp=sharing.

----

## Proposed change

Added a one-hour preflight cache duration to Hubble's global CORS configuration.

### DIFF
--- a/hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/config/GlobalCorsConfig.java
+++ b/hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/config/GlobalCorsConfig.java
@@ -35,6 +35,7 @@ public class GlobalCorsConfig {
         config.setAllowCredentials(false);
         config.addAllowedMethod("*");
         config.addAllowedHeader("*");
+        config.setMaxAge(3600L);
 
         UrlBasedCorsConfigurationSource source =
                 new UrlBasedCorsConfigurationSource();


imbajin (commented):
Blocking: yes. Summary: The configured preflight cache is unreachable under the current same-origin-only policy, so this change cannot produce the advertised response header. Evidence: Spring Web 5.1.9 rejects an empty allowedOrigins configuration before writing maxAge; the Hubble backend build, Checkstyle, and 394 unit tests pass.

hugegraph-hubble/hubble-be/src/main/java/org/apache/hugegraph/config/GlobalCorsConfig.java
```diff
@@ -35,6 +35,7 @@ public CorsFilter corsFilter() {
         config.setAllowCredentials(false);
         config.addAllowedMethod("*");
         config.addAllowedHeader("*");
+        config.setMaxAge(3600L);
```
imbajin, line 38: ‼️ **Critical: the configured preflight cache is unreachable.** This `CorsConfiguration` never sets `allowedOrigins`; in the pinned Spring Web 5.1.9, `checkOrigin()` returns `null` for an empty origin list and `DefaultCorsProcessor` rejects the preflight before it reaches the branch that writes `Access-Control-Max-Age`. Same-origin requests do not need CORS preflight processing, so this line cannot implement the PR's stated behavior. Please either remove this ineffective setting while Hubble remains same-origin-only, or first add an explicit trusted-origin configuration and an `OPTIONS` regression test proving that an allowed origin receives `Access-Control-Max-Age: 3600` while untrusted origins remain rejected.

zyxxoo (approved):

imbajin (changes requested):

PR #760, opened 2026-08-23
fix(client): support graph-scoped legacy user auth paths

### LINKED ISSUE #759: [Bug] Client UserAPI falls back to an incompatible auth path on Server 1.5
### Bug Type (问题类型)

rest-api (结果不合预期)

### Before submit

- [x] I had searched in the [issues](https://github.com/apache/hugegraph-toolchain/issues) and found no similar issues.

### Environment (环境信息)

- HugeGraph Server: 1.5.0 (Core 1.5.0, REST API 0.71.0.0)
- Reproduction Hubble/Toolchain baseline: 60df3c51 (before the compatibility changes)
- Current upstream PR target: be7ef3ae
- Authentication: StandardAuthenticator enabled
- Deployment: standalone, DEFAULT / hugegraph
- OS: macOS
- Browser: Google Chrome
- Data size: not data-dependent

### Expected & Actual behavior (期望与实际表现)

When the current Java Client connects to an authentication-enabled HugeGraph Server 1.5, `AuthManager` creates `UserAPI` with the GraphSpace-scoped path. Because Server 1.5 does not support GraphSpace, `RestClient` removes the `graphspaces/DEFAULT/` prefix and sends the request to `/auth/users`. Server 1.5 instead requires `/graphs/{graph}/auth/users`, and that legacy endpoint returns a `users` wrapper.

As a result, user lookup fails even when the supplied Basic credentials are valid. The Hubble screenshots below show the broader user-visible login symptom; this issue is intentionally scoped to the Java Client `UserAPI` path and response handling. Full Hubble login compatibility should be handled separately.

## Steps to reproduce

1. Start HugeGraph Server 1.5.0 with `StandardAuthenticator` enabled.
2. Confirm `GET /versions` reports Core 1.5.0.
3. Confirm the test credentials are accepted by `GET /graphs/hugegraph/auth/users?name=<user>`.
4. Create a current pre-fix `HugeClient` for `DEFAULT / hugegraph`.
5. Call `client.auth().getUser(<existing-user>)`.
6. Observe that the client falls back to `/auth/users` instead of `/graphs/hugegraph/auth/users`, so the lookup fails.

### Step 1: Open the Hubble login page

Hubble login page before compatibility fix

### Step 2: Enter credentials already verified against the Server 1.5 graph-scoped auth endpoint

Verified local credentials entered with password masked

### Step 3: Submit the login form

Hubble login request fails against Server 1.5

## Actual behavior

The legacy server request is routed to the unscoped `/auth/users` endpoint and user lookup fails. In Hubble this surfaces as `Request failed: Failed to login HugeGraph Server`.

## Expected behavior

When GraphSpace is unsupported and a graph name is available, `UserAPI` should use `/graphs/{graph}/auth/users` and read the legacy `users` response wrapper. The existing GraphSpace-scoped behavior for modern servers should remain unchanged.

## Proposed focused fix

- Add a graph-aware legacy auth path in the Java Client.
- Use it only when GraphSpace is unsupported and a graph name is present.
- Parse the Server 1.5 `users` wrapper in `UserAPI.getByName()`.
- Add regression tests for both legacy and modern paths.

```javascript
N/A — this issue is not dependent on vertex or edge data.
```

```javascript
N/A — this issue is caused by authentication endpoint compatibility, not graph schema.
```

looksaw2:
## Summary

- use `graphs/{graph}/auth/users` when the server does not support GraphSpace;
- request all legacy users with `limit=-1`, then find the exact username in the Server 1.5 `users` wrapper;
- preserve the existing GraphSpace-scoped path and `name` query for modern servers;
- add regression coverage for the legacy path, a target user after the default first 100 records, and the unchanged modern path.

## Environment and reproduction

- HugeGraph Server: 1.5.0, Core 1.5.0, REST API 0.71.0.0
- Authentication: `StandardAuthenticator`
- Context: `DEFAULT / hugegraph`
- PR base: `apache/hugegraph-toolchain@be7ef3ae`
- OS: macOS

The full Before flow and three step-by-step screenshots are recorded in #759.

## After verification

1. Real Server 1.5 contract:
   - `GET /graphs/hugegraph/auth/users?limit=-1` -> HTTP 200
   - `GET /auth/users?limit=-1` -> HTTP 404
   - response wrapper -> `users`
2. Patched Java Client against the real Server 1.5 runtime:
   - `client.auth().getUserByName("admin")` -> `user=admin success=true`
3. Regression suite:
   - 63 tests, 0 failures, 0 errors
   - 0 Checkstyle violations

### Step 1: verify the real Server 1.5 contract

Server 1.5 legacy auth contract

### Step 2: run the patched Java Client against Server 1.5

Patched client user lookup succeeds

### Step 3: run focused and unit regression tests

Client regression tests pass

## Tests

```bash
cd hugegraph-client
mvn test -Dtest=AuthApiPathTest,UnitTestSuite \
  -Dmaven.javadoc.skip=true -ntp
```

Fixes #759

### DIFF
--- a/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
+++ b/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
@@ -30,8 +30,17 @@ import com.google.common.collect.ImmutableMap;
 
 public class UserAPI extends AuthAPI {
 
+    private final boolean legacyGraphScoped;
+
     public UserAPI(RestClient client, String graphSpace) {
         super(client, graphSpace);
+        this.legacyGraphScoped = false;
+    }
+
+    public UserAPI(RestClient client, String graphSpace, String graph) {
+        super(client, graphSpace, graph);
+        this.legacyGraphScoped = !client.isSupportGs() &&
+                                 graph != null && !graph.isEmpty();
     }
 
     @Override
@@ -70,8 +79,19 @@ public class UserAPI extends AuthAPI {
     }
 
     public User getByName(String name) {
-        Map<String, Object> params = ImmutableMap.of("name", name);
+        Map<String, Object> params = this.legacyGraphScoped ?
+                                     ImmutableMap.of("limit", -1) :
+                                     ImmutableMap.of("name", name);
         RestResult result = this.client.get(this.path(), params);
+        if (this.legacyGraphScoped) {
+            List<User> users = result.readList(this.type(), User.class);
+            for (User user : users) {
+                if (name.equals(user.name())) {
+                    return user;
+                }
+            }
+            return null;
+        }
         return result.readObject(User.class);
     }
 
--- a/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
+++ b/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.hugegraph.client.RestClient;
+import org.apache.hugegraph.rest.ClientException;
 import org.apache.hugegraph.rest.RestResult;
 import org.apache.hugegraph.structure.auth.User;
 import org.apache.hugegraph.structure.auth.User.UserRole;
@@ -30,17 +31,12 @@ import com.google.common.collect.ImmutableMap;
 
 public class UserAPI extends AuthAPI {
 
-    private final boolean legacyGraphScoped;
-
     public UserAPI(RestClient client, String graphSpace) {
         super(client, graphSpace);
-        this.legacyGraphScoped = false;
     }
 
     public UserAPI(RestClient client, String graphSpace, String graph) {
         super(client, graphSpace, graph);
-        this.legacyGraphScoped = !client.isSupportGs() &&
-                                 graph != null && !graph.isEmpty();
     }
 
     @Override
@@ -79,19 +75,17 @@ public class UserAPI extends AuthAPI {
     }
 
     public User getByName(String name) {
-        Map<String, Object> params = this.legacyGraphScoped ?
-                                     ImmutableMap.of("limit", -1) :
-                                     ImmutableMap.of("name", name);
-        RestResult result = this.client.get(this.path(), params);
-        if (this.legacyGraphScoped) {
-            List<User> users = result.readList(this.type(), User.class);
+        if (this.legacyGraphScoped()) {
+            List<User> users = this.list(-1);
             for (User user : users) {
                 if (name.equals(user.name())) {
                     return user;
                 }
             }
-            return null;
+            throw new ClientException("User '%s' does not exist", name);
         }
+        Map<String, Object> params = ImmutableMap.of("name", name);
+        RestResult result = this.client.get(this.path(), params);
         return result.readObject(User.class);
     }
 
--- a/hugegraph-client/src/main/java/org/apache/hugegraph/driver/AuthManager.java
+++ b/hugegraph-client/src/main/java/org/apache/hugegraph/driver/AuthManager.java
@@ -66,7 +66,7 @@ public class AuthManager {
         this.targetAPI = new TargetAPI(client, graphSpace);
         this.groupAPI = new GroupAPI(client);
         this.graphSpaceGroupAPI = new GroupAPI(client, graphSpace);
-        this.userAPI = new UserAPI(client, graphSpace);
+        this.userAPI = new UserAPI(client, graphSpace, graph);
         this.accessAPI = new AccessAPI(client, graphSpace);
         this.projectAPI = new ProjectAPI(client, graphSpace);
         this.belongAPI = new BelongAPI(client, graphSpace);
--- a/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/AuthAPI.java
+++ b/hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/AuthAPI.java
@@ -27,23 +27,33 @@ public abstract class AuthAPI extends API {
     private static final String LEGACY_PATH = "graphs/%s/auth/%s";
     private static final String USER_PATH = "auth/%s";
 
+    private final boolean legacyGraphScoped;
+
     public AuthAPI(RestClient client) {
         super(client);
+        this.legacyGraphScoped = false;
         this.path(USER_PATH, this.type());
     }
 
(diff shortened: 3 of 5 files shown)


bitflicker64 (commented):
Blocking: no. Summary: The graph-scoped routing and the `limit=-1` scan match the server 1.5.0 contract, and the GraphSpace path is unchanged for modern servers. One asymmetry is worth closing first: `getByName` now returns `null` for a missing user, an outcome the GraphSpace branch never produces, and hubble's callers are split on whether they guard it. Evidence: `git diff be7ef3ae..be1df870` over the exact head; `mvn -o test -Dtest=AuthApiPathTest` in `hugegraph-client` on JDK 11 passes 3/3; `apache/hugegraph` at tag 1.5.0 has `@Path("graphs/{graph}/auth/users")` whose `list` accepts only `limit`, and `graphs/{graph}/auth/{targets,groups,accesses,belongs,projects}` for the siblings; `readObject(User.class)` on a `{"users":[...]}` wrapper throws `SerializeException`, reproduced against the built client classes; every workflow run on this head is `action_required`, so there is no CI signal.

hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
```diff
@@ -70,8 +79,19 @@ public List<User> list(int limit) {
...
+                                     ImmutableMap.of("name", name);
         RestResult result = this.client.get(this.path(), params);
+        if (this.legacyGraphScoped) {
+            List<User> users = result.readList(this.type(), User.class);
+            for (User user : users) {
+                if (name.equals(user.name())) {
+                    return user;
+                }
+            }
+            return null;
```
bitflicker64: ⚠️ The legacy branch returns `null` on a miss, a value the GraphSpace branch never produces: there `readObject(User.class)` either yields a user or throws, so `null` is a new third outcome for the same method.

Hubble's callers are split on whether they handle it. `UserService.getpersonal` guards it (line 172), while `updatePersonal` (line 406) and `updatepwd` (line 433) dereference the result straight away, and `getUser` (line 100) turns a miss into a null entity rather than an error.

Please throw on a miss so both branches signal it the same way. Two cheap `AuthApiPathTest` additions would pin this down: a name absent from the response, and `graph == null` with `supportGs` false, which is the fall-through `ManagerAPITest` already exercises via `new AuthManager(client, "DEFAULT", null)`.

hugegraph-client/src/main/java/org/apache/hugegraph/driver/AuthManager.java
```diff
@@ -66,7 +66,7 @@ public AuthManager(RestClient client, String graphSpace, String graph) {
         this.targetAPI = new TargetAPI(client, graphSpace);
         this.groupAPI = new GroupAPI(client);
         this.graphSpaceGroupAPI = new GroupAPI(client, graphSpace);
-        this.userAPI = new UserAPI(client, graphSpace);
+        this.userAPI = new UserAPI(client, graphSpace, graph);
```
bitflicker64, line 69: 🧹 Follow-up note rather than a request to widen this PR: #759 scopes the fix to `UserAPI`, and the other auth entities were already broken against 1.5. Worth capturing before it is lost, since "Fixes #759" closes the tracking issue on merge.

`targetAPI`, `graphSpaceGroupAPI`, `accessAPI`, `projectAPI`, `belongAPI` and `managerAPI` still take the two-argument constructor, so against a 1.5 server they build `graphspaces/DEFAULT/auth/<type>`, which `RestClient.removeDefaultGsPrefix` rewrites to `auth/<type>`. Server 1.5.0 has no such routes: `TargetAPI` is `@Path("graphs/{graph}/auth/targets")`, and `GroupAPI`, `AccessAPI`, `BelongAPI` and `ProjectAPI` are graph-scoped the same way. Hubble's `RoleService`, `BelongService`, `AccessService` and `TargetService` all depend on them.

Could you open a follow-up issue for the remaining auth surface and link it here?

hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
```diff
@@ -30,8 +30,17 @@
...
+    private final boolean legacyGraphScoped;
+
     public UserAPI(RestClient client, String graphSpace) {
         super(client, graphSpace);
+        this.legacyGraphScoped = false;
+    }
+
+    public UserAPI(RestClient client, String graphSpace, String graph) {
+        super(client, graphSpace, graph);
+        this.legacyGraphScoped = !client.isSupportGs() &&
```
bitflicker64: 🧹 This repeats the predicate `AuthAPI` already evaluated to pick between `LEGACY_PATH` and `PATH`, with nothing tying the two copies together. If a later edit moves one, `getByName` sends the wrong query for the path it is actually on: `name=<n>` against `graphs/{graph}/auth/users` is ignored by server 1.5.0, which returns the first 100 users in a `users` wrapper, and `readObject(User.class)` then fails with `SerializeException: Failed to deserialize` (reproduced against the built client classes). That is the exact symptom #759 reports.

Suggest storing the decision once in `AuthAPI` (a `protected final boolean` plus an accessor) and reading it here. The legacy branch of `getByName` is `list(-1)` inlined, so it can call `this.list(-1)` and scan the result, which also restores the `checkLimit` guard the hand-built params map skips.

hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
imbajin: Fixed in a12819b2. Legacy misses now throw a clear ClientException, and the regression coverage includes both an absent username and the graph-null fallback.

hugegraph-client/src/main/java/org/apache/hugegraph/driver/AuthManager.java
imbajin: Confirmed as a broader pre-existing legacy-auth gap, but it is intentionally outside this UserAPI-focused PR. Hubble 1.5 support is limited to non-auth standalone mode, so no follow-up is needed unless legacy authenticated mode becomes a supported target.

hugegraph-client/src/main/java/org/apache/hugegraph/api/auth/UserAPI.java
imbajin: Fixed in a12819b2. AuthAPI now owns the legacy-path decision, UserAPI reads that single source, and the legacy lookup reuses list(-1).

imbajin (approved):

